PR #8, opened 2015-03-11
General cleanups and add link to minibuffer on kill

apg:
Biggest change is that I added the link to the minibuffer on kill so I can know it did something. Otherwise, just some cleanups around indentation, and removed a cleanup that you had regarding functionp???? Basically, since elisp is a Lisp 2, symbols have values attached to them, so `(symbol-function 'symbol)` will get you the function, or it'll be nil. Also, moved to `when` for one armed `if` statements.

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -92,20 +92,20 @@
 
 (defun git-link-relative-filename ()
   (let* ((filename (buffer-file-name))
-	 (dir      (git-link-repo-root)))
+         (dir      (git-link-repo-root)))
     (if (and dir filename)
-	(substring (file-truename filename)
-		   (1+ (length dir))))))
+        (substring (file-truename filename)
+                   (1+ (length dir))))))
 
 (defun git-link-remote-host (remote-name)
   (let ((url (git-link-remote-url remote-name)))
     (if (string-match git-link-remote-regex url)
-	(match-string 1 url))))
+        (match-string 1 url))))
 
 (defun git-link-remote-dir (remote-name)
   (let ((url (git-link-remote-url remote-name)))
     (if (string-match git-link-remote-regex url)
-	(match-string 2 url))))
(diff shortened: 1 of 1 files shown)


apg:
will conflict with issue 10 pull request, which sucks. will rebase and fix after issue 10 is applied (assuming it gets applied)

sshaw:
Good idea to add the URL to the mini buffer. Give me a few to look at all these, though some of these seem ideological :neckbeard::

1. `functionp` -while not used- seems a bit more idiomatic than `symbol-function`.  `symbol-function` sort of feels like doing `int.to_s.length < 2` instead of just saying `int < 10`
2. `when` calls `if`, `if`'s else is optional; what else are the difference and upside to the change?
3. What was the problem with the indentation?

apg:
>  functionp -while not used- seems a bit more idiomatic than symbol-function. symbol-function sort of feels like doing int.to_s.length < 2 instead of just saying int < 10 

Yes, functionp is the better choice here. Sorry, it seemed as though, from the comment, you tried it and it didn't work. I'll update this pull request with a call to functionp instead.

> when calls if, if's else is optional; what else are the difference and upside to the change?

Yes, `if` works the way you describe, but when is more readable when `if` is one armed, because it doesn't. `when` is just a macro that'll expand into the one armed `if`, so there's no runtime penalty, but readability is improved (in my honest opinion)

> What was the problem with the indentation?

Just non-standard. Parens in lets and stuff that shouldn't align that did. I just reindented the buffer.

apg:
Actually, `functionp` *doesn't* work there because handler is a symbol, not a function object.

apg:
well, it does if you use 'handler, pull-request updated.

sshaw:
@apg #10 has been applied

Issue #12, opened 2015-03-17
git-link-commit in master 


kaushalmodi:
Hi,

When I try to select a region and call `git-link-commit`, I get the error

> Point is not on a commit hash

Instead of that, can you return the link to the selected lines from the **latest** commit?

Most of the time, I need to send reference to a piece of code from my master (latest commit). But that code might change in the future and the line references might change. So I prefer to link to the links in that specific commit even if that's the master.

sshaw:
Resolved by a570765.

PR #9, opened 2015-03-12
Utilize completing-read for remote prompt.

apg:
- Cleanup usage of interactive and handling region, prefix args.

PR #13, opened 2015-03-17
Added an option to force using commit hash + other stuff

kaushalmodi:
- By default, the branch name is used in the copied link if available
- The `defvar` `git-link-force-commit-hash-in-link` when set to `t`
  forces the use of the commit hash in the copied link (even if a branch
  name like `master` is available.
- The interactive function `git-link-force-hash` set this var to true in
  a `let` before calling `git-link`.
- This commit does NOT change the default functionality of the package.

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -142,10 +142,26 @@
   (save-restriction
     (widen)
     (save-excursion
-      (let ((use-region (use-region-p)))
-        (list
-         (line-number-at-pos (when use-region (region-beginning)))
-         (when use-region (line-number-at-pos (region-end))))))))
+      (let* ((use-region (use-region-p))
+             (start (when use-region (region-beginning)))
+             (end   (when use-region (region-end)))
+             (line-start (line-number-at-pos start))
+             line-end)
+        (when use-region
+          ;; When region is selected from bottom to top, exchange point and mark
+          ;; so that the `point' and `(region-end)' are the same
+          (when (< (point) (mark))
+            (exchange-point-and-mark))
+          ;; If the `end' position is at the beginning of a line
+          ;; decrement the position by 1, so that the resultant
+          ;; position is on the previous line.
+          (when (= end (line-beginning-position))
+            (setq end (1- end)))
+          (setq line-end (line-number-at-pos end))
+          ;; Set the `line-end' to `nil' if `line-end' <= `line-start'
+          (when (<= line-end line-start)
+            (setq line-end nil)))
+        (list line-start line-end)))))
 
 (defun git-link-github (hostname dirname filename branch commit start end)
   (format "https://%s/%s/blob/%s/%s#%s"
--- a/git-link.el
+++ b/git-link.el
@@ -195,6 +195,7 @@
 
 (defun git-link-new (link)
   (message (kill-new link))
+  (setq deactivate-mark t)
   (when git-link-open-in-browser
     (browse-url link)))
 


kaushalmodi:
Solution for Issue https://github.com/sshaw/git-link/issues/12

sshaw:
Hi, thanks for this. To clarify: If point is not on a commit hash, you want it to default to generating a link to the commit for the given line/region instead of the current branch? If so I like that idea, and find myself doing the same thing for the same reason when creating links in GitHub issues  (though there I mostly use tags). 

It may be better to make this part of `git-link` and enable it via a prefix arg or a `defvar`. No need for a separate function. Ultimately I'd like to have one core function: `git-link-dwim`. 

The variable name `git-link-force-commit-hash-in-link` is a bit verbose. Maybe `git-link-use-commit` is better? Thoughts?

I do need to sort out #8, #9, and #11 first.

kaushalmodi:
>  If point is not on a commit hash, you want it to default to generating a link to the commit for the given line/region instead of the current branch? 
Yes

> It may be better to make this part of git-link and enable it via a prefix arg or a defvar. No need for a separate function. 
You were already using prefix arg for `prompt`. So I did not want to disturb that. You might be open to supporting multiple values of the prefix arg: `C-u` -> 4, `C-u C-u` -> 16, ...

> Ultimately I'd like to have one core function: git-link-dwim.
That would be great.

> The variable name git-link-force-commit-hash-in-link is a bit verbose. Maybe git-link-use-commit is better? Thoughts?
Agreed.

> I do need to sort out #8, #9, and #11 first.
Once those requests are sorted out, I can continue on improving this pull request.

Thanks!

sshaw:
Also, pull in latest master and squash future commit. Thanks.

kaushalmodi:
Thanks. I will take out some time by next week. I have not used git for anything much else than commits to my .emacs.d. So I'll need time to figure out how to do what you suggested.

Basically I need to rebase my fork to your latest version first and replace this pull request with the one with my changes on top of the rebased fork? But I need to figure out how to do that.

sshaw:
Actually based on [the latest observations](https://github.com/sshaw/git-link/pull/13#commitcomment-10675837) I would just toss this and start anew based on master.

kaushalmodi:
Can you please review this? If this is clean, then we can find a solution on how not to have a separate `git-link-force-hash` defun.

sshaw:
Hi @kaushalmodi, I missed some of the other commits earlier. 

Can you just submit the `git-link-use-commit` stuff as now it's mixed with the [`deactivate-mark`](https://github.com/kaushalmodi/git-link/commit/dd1ace5e66621995e6aedd6f1539206541d3fc9a) and [line number fixes](https://github.com/kaushalmodi/git-link/commit/f2d72970808bd4f4fcc28d42b861c5d205f0295b). 

It's easy for me to tackle one issue at a time.

kaushalmodi:
The changes don't overlap and not many lines have changed. Can you please review these this time?

Also, how do I control what gets appended to the pull request?

kaushalmodi:
Cleaned up the commits a bit. But I still can't figure out how to gate the later commits.

sshaw:
> The changes don't overlap and not many lines have changed. Can you please review these this time?

Yes, I've pulled in e560b41.

> Also, how do I control what gets appended to the pull request?

Anything on the branch for the pull request will be added. If there's something unrelated to the request just create a new branch and commit the work there. In general it's good to stick to one feature or bug fix per request. 

As for 717f46f this is caused by [your emacs config](https://github.com/kaushalmodi/.emacs.d/blob/master/setup-files/setup-misc.el#L18-24). But, this is something that should be resolved. I'm just not sure if 717f46f is the right way. I'm open to other suggestions but will have to do some research myself. 

I'll also have a look at aee3689.

Thanks again!

kaushalmodi:
> Yes, I've pulled in e560b41.

Thanks!

> If there's something unrelated to the request just create a new branch and commit the work there. In general it's good to stick to one feature or bug fix per request.

Thanks, will do so now on.

> As for 717f46f this is caused by your emacs config. 

Wow! Thanks for culling that from my config.

> I'm just not sure if 717f46f is the right way. I'm open to other suggestions but will have to do some research myself.

Once you are done copying the git-link, the region selection shouldn't be needed. Is that a correct assumption? If so, deactivating mark would be harmless. I have needed to do the same in a [different function too](https://github.com/kaushalmodi/.emacs.d/tree/a09ecb680a3073e6cfaabb264382d3d3f3da8787/setup-files/setup-editing.el#L224). 

> I'll also have a look at aee3689.

Thanks!

sshaw:
> Once you are done copying the git-link, the region selection shouldn't be needed. Is that a correct assumption? 

I think so.

> If so, deactivating mark would be harmless. 

After some research I think it's better to go with `(setq deactivate-mark t)` over `(deactivate-mark)`.

* `(deactivate-mark)` does a bunch of other things, from <kbd>C-h f</kbd> `deactivate-mark`:

```text
Deactivate the mark.
If Transient Mark mode is disabled, this function normally does
nothing; but if FORCE is non-nil, it deactivates the mark anyway.

Deactivating the mark sets `mark-active' to nil, updates the
primary selection according to `select-active-regions', and runs
`deactivate-mark-hook'.

If Transient Mark mode was temporarily enabled, reset the value
of the variable `transient-mark-mode'; if this causes Transient
Mark mode to be disabled, don't change `mark-active' to nil or
run `deactivate-mark-hook'.
```

My conclusion here: side effects are bad, especially if you don't know how they will manifest themselves.  Of course my reasoning here could be seen as a side effect of my naïveté.

* `kill-region` uses `(setq deactivate-mark t)`

What do you think?

> I have needed to do the same in a different function too. Well, the source from which that function was derived already had that deactivate-mark line.

[That is a very nice function].(https://github.com/kaushalmodi/.emacs.d/blob/a09ecb680a3073e6cfaabb264382d3d3f3da8787/setup-files/setup-editing.el#L224).

kaushalmodi:
Here's a rough survey of how many times the function `deactivate-mark` is used vs just the setq.

It looks like its use as function is more prominent. But using simply `(setq deactivate-mark t)` worked for me too. So I will update the pull request with that.

But now I am curious to know when one should be used vs the other.

**Used as a function:**
```
km²~/downloads/:git/emacs> ag '\(deactivate-mark[\st]*\)' -G '\.el' -c
emacs-lisp/ert-x.el:1
emulation/cua-base.el:1
emulation/edt.el:1
emulation/viper-util.el:1
erc/erc-goodies.el:2
emacs-lisp/tcover-ses.el:1
gnus/nnnil.el:1
isearch.el:1
info.el:2
international/mule.el:2
leim/quail/hangul.el:1
mail/rmail.el:1
mwheel.el:1
replace.el:2
mouse.el:5
obsolete/mouse-sel.el:1
obsolete/tpu-edt.el:1
org/org-w3m.el:1
ses.el:1
ps-print.el:1
org/org.el:4
progmodes/f90.el:2
skeleton.el:1
tempo.el:1
term/ns-win.el:2
startup.el:1
term.el:1
textmodes/picture.el:2
simple.el:10
textmodes/reftex-index.el:1
textmodes/bibtex.el:1
type-break.el:1
vc/ediff-util.el:1
winner.el:1
```

**Only setting the var:**
```
km²~/downloads/:git/emacs> ag '\(setq deactivate-mark t\)' -G '\.el' -c
delsel.el:1
emulation/cua-rect.el:1
emulation/cua-base.el:1
indent.el:1
ses.el:2
rect.el:2
register.el:4
simple.el:2
```

sshaw:
Interesting to see how the usage of the two are distributed. In the case of `(deactivate-mark)` while to they look to all be on line one? Maybe I'm not understanding the output?

kaushalmodi:
The number following the colon is the number of occurrences in that file. It is because I use the `-c` argument for `ag`.

I learn more, I have [posted this on emacs.SE](http://emacs.stackexchange.com/q/12394/115).

kaushalmodi:
Updated the pull req.

kaushalmodi:
Thanks! I have squashed the bug-fix commit into the previous one and tested this one to work fine with and without region selections.

git-link.el
```diff
@@ -142,10 +142,22 @@
...
+      (let* ((use-region (use-region-p))
+             (start (when use-region (region-beginning)))
+             (end   (when use-region (region-end)))
+             (line-start (line-number-at-pos start))
+             line-end)
+        (when use-region
+          ;; If the `end' position is at the beginning of a line
+          ;; decrement the position by 1, so that the resultant
+          ;; position is on the previous line
+          (when (= end (line-beginning-position))
```
sshaw: This does not account for the reverse scenario: the point position is less than the mark position. In this case the extraneous line is still included. For example, `*P*` is point, `*M*` is mark:

```
1. *P*one
2. two
3. three
4. *M*
```
calling `git-link` with the arraignment will include lines 1-4.  

Need to check if `(region-end)` is at the beginning of the line, not `(point)`. Off hand not sure how to do this without moving point. Ideas?

kaushalmodi: I believe I had tested that scenario and also had read that
region-beginning and region-end automatically swap the start and end
positions such that end is greater than or equal to start.

I'll test that scenario when I get to a computer.

sshaw: The key is that `line-beginning-position` is compared to the value given by `region-end` (`end`).
Since `line-beginning-position` will use `point`, if `point` is above mark it's not at the end of the region and the comparison returns `nil`.

kaushalmodi:
I first got the solution with this convoluted logic, 

``` el
          (when (or
                 ;; when region is selected from top to bottom
                 (and (> (point) (mark))
                      (= end (line-beginning-position)))
                 ;; when region is selected from bottom to top
                 (and (< (point) (mark))
                      (= start (line-beginning-position))))
            (setq end (1- end)))
```

but then came up with a better solution; squashed and pushed the commit.

The point is not affected as stuff happens in save-excursion.

kaushalmodi:
Thanks. This pull request was a great learning experience.

sshaw:
Thanks. 

So seems like there's no way to get the beginning of a line's position without moving point huh?

kaushalmodi:
There is.. but by using the `(when (or...)) ` logic in the above comment,  which is a bit convoluted compared to the current solution. Also the point is unaffected in the current solution as we use `save-excursion`.

PR #15, opened 2015-04-05
Use remote hostname when constructing link URLs.

dhull:
GitHub, Gitorious, and Bitbucket each have self-hosted versions (GitHub
Enterprise, Gitorious open source, and Atlassian Stash).  Update the
link-creation functions so that they use the remote hostname to contruct
the link URL instead of hardcoding each company's hostname so that the
link functions can be used to contruct URLs for the self-hosted versions
of the products.


sshaw:
Great, thanks!

Issue #17, opened 2015-05-13
Configuring remote per repository


prathamesh-sonpatki:
Can we configure default remote (origin/upstream) (& branch also) at repository level. For certain upstream repositories, I want to see commit links in upstream though I am in origin. Usecase for forked codebases. Upstream is pointing to source while origin is pointing to my fork which may be outdated.

sshaw:
Sounds like a good idea. Currently you may be able to do this for remote by advising `git-link` (untested):
```elisp
(defadvice git-link (aroud git-link-remote-lookup first nil activate)
  (let ((git-link-default-remote (your-lookup-funtion (git-link-repo-root))))
    ad-do-it))
```
But that doesn't handle branch, which currently can't be set. 

How would you expect to map repo to branch/remote? Per buffer via buffer local variables, a callback, prompted by prefix call to `git-link`? What would be the key?

sshaw:
> Can we configure default remote (origin/upstream) (& branch also) at repository level

Ah, I think you're saying that you'd like to be able so do something like this:

```
git config --local --add git-link.remote  some-remote
git config --local --add git-link.branch  some-branch
```

Good idea. Please confirm.

prathamesh-sonpatki:
@sshaw Sorry for replying late. But this is exactly I want :smile:

sshaw:
@prathamesh-sonpatki checkout the [git-based-config branch](https://github.com/sshaw/git-link/tree/git-based-config). It gives priority to the the branch and remote given by `git config git-link.remote` and `git config git-link.branch`. Let me know what you think.

One question: If `git-link.remote` does not exist how should <kbd>C-u M-x</kbd> <kbd>git-link</kbd>
behave? I think current functionality is okay, but I can also see it being a tad confusing.

sshaw:
This has been merged into master and can be found in the `v0.3.0` release.

prathamesh-sonpatki:
Looks good. I will try it out. 
 
> If git-link.remote does not exist how should C-u M-x git-link
behave?

I think current behavior is good in that case.

PR #18, opened 2015-05-28
Add optional remote-host mapping

dougm:
When using gerrit, the git remote url looks something like:
 "ssh://username@gerrit.example.com:29418/project-name"

It is common to have gerrit replicate to github or gitlab, this change
makes it possible to add this mapping from gerrit host to another such
service.  For example, mapping the above url to:
 "https://gitlab.example.com/org-name/project-name"

Changed the url regex to use url-parse, since the port was being
included as part of the path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,18 @@ The `git-link-commit` signature is:
 * `DIRNAME` directory portion of the remote
 * `COMMIT` SHA of the commit
 
+If you use Gerrit with replication enabled to a supported service,
+you can redirect by adding the following to your `.emacs` file.
+
+    (eval-after-load "git-link"
+      '(progn
+        (add-to-list 'git-link-remote-host-alist
+          '("gerrit.example.com" "gitlab.example.com/org-name"))
+        (add-to-list 'git-link-remote-alist
+          '("gitlab.example.com/org-name" git-link-github))
+        (add-to-list 'git-link-commit-remote-alist
+          '("gitlab.example.com/org-name" git-link-commit-github))))
+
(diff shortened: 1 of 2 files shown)


sshaw:
I believe I looked into `url-generic-parse-url` and didn't use it because it can't parse scp style URLs; the ones like `git` uses:

```
ELISP> (url-generic-parse-url "http://a.com:5000/a/path")
[cl-struct-url "http" nil nil "a.com" 5000 "/a/path" nil nil t nil t]
ELISP> (url-generic-parse-url "git@github.com:sshaw/tidy_gemfile.git")
[cl-struct-url nil nil nil nil nil "git@github.com:sshaw/tidy_gemfile.git" nil nil nil nil t]
```

That aside, I do have some concerns about adding another ` git-link-*-alist` variable. I don't really like the current paradigm: one alist for commits and one for general linking. I think the 2 should be consolidated -somehow. 

In your example above `ssh://username@gerrit.example.com:29418/project-name` is a returned by
 `git config --get remote.xxxx.url`?

What if you created a dummy remote `https://gitlab.example.com/org-name/project-name` then set `git-link-default-remote` to it?

sshaw:
Actually, instead of setting up a dummy remote, what about [something like this](https://github.com/sshaw/git-link/issues/17#issuecomment-109139444)?

dougm:
Ah, yes, adding gitlab as a remote + git-link-default-remote works great.  #17 sounds useful too.  Agreed, no need for this patch then.  Thanks!

PR #19, opened 2015-06-04
git-link-new errors on call to message if interprogram-cut-function is set

rlister:
If `interprogram-cut-function` is set (for example, OSX users often set this to invoke `pbcopy`) `kill-new` may return a proc instead of a string, causing `message` to error. Calling `message` directly on `link` should be safer.

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -212,7 +212,8 @@
 	  commit))
 
 (defun git-link-new (link)
-  (message (kill-new link))
+  (kill-new link)
+  (message link)
   (setq deactivate-mark t)
   (when git-link-open-in-browser
     (browse-url link)))


sshaw:
Hi thanks for this. 

What function causes this for you? On my OS X setup `interprogram-cut-function` is set to `x-select-text` and I don't have this problem.  
Are you using Aquamacs?

rlister:
Using: `GNU Emacs 24.5.1 (x86_64-apple-darwin14.3.0, NS apple-appkit-1347.57) of 2015-05-05`.

I have the following setup:

```lisp
(defun cut-to-osx (text &optional push)
  (let ((process-connection-type nil))
    (let ((proc (start-process "pbcopy" "*Messages*" "reattach-to-user-namespace" "pbcopy") ))
      (process-send-string proc text)
      (process-send-eof proc))))

(setq interprogram-cut-function 'cut-to-osx)
```

I guess my problem is that `process-send-eof` returns the process it is called on (in this case `#<process pbcopy<1>>`). In retrospect I could have fixed this by returning `text` from my `defun`. However, all things considered, it's safer that we don't assume `kill-new` returns a string, considering some of the clipboard hackage that exists in the wild :smile: 

Thanks!

PR #21, opened 2015-07-23
GitHub fix

carlesso:
Hi guys! Finally I found what I was looking for. I've a custom gitlab repository and now finally I can get the link to the repo!

I've noticed that when selecting a region the generated link was wrong, it was using the github format: `#L1-L24` while gitlab does not want the last L, so the correct format is `#L1-24`. So here a pull request for you! :)


sshaw:
Thanks for the commit. This has been released with `v0.2.1`.

carlesso:
That's great! Thanks.

PR #28, opened 2016-02-17
Support git savannah gnu org

kaushalmodi:


sshaw:
Closing as #29 contains the same commits.

PR #29, opened 2016-02-22
Support orgmode org

kaushalmodi:
Support the org-mode source code git

The git-link-remote-regex had to be fixed to support links like
http://orgmode.org/org-mode.git

Note that earlier regexp expected 2 forward slashes in the path, like
http://git.savannah.gnu.org/r/emacs.git

Now the second forward slash part is put into an optional regexp
sub-group "\(/[^/]+?\)*"

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -90,7 +90,8 @@ word under point is not a valid commit hash.")
     ("bitbucket.org"        git-link-bitbucket)
     ("gitorious.org"        git-link-gitorious)
     ("gitlab.com"           git-link-gitlab)
-    ("git.savannah.gnu.org" git-link-savannah-gnu))
+    ("git.savannah.gnu.org" git-link-savannah-gnu)
+    ("orgmode.org"          git-link-org-mode))
   "Maps remote hostnames to a function capable of creating the appropriate file URL")
 
 (defvar git-link-commit-remote-alist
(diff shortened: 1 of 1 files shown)


sshaw:
Hi, thanks for another pull request. 

I don't think it makes sense to add support for linking to a specific project's repository (i.e., org-mode).  `git-link-remote-alist` and `git-link-commit-remote-alist` are exposed so that one can link to stuff like this. If they're not flexible enough to support linking to certain repos then the pull request should specifically target a fix for them. 

Savannah sorta makes sense, as it *can be* a git service but **if** the project owner choses to use git, but [Savannah's own statistics](https://savannah.gnu.org/stats/#between) seem to indicate that -short of some popular, venerable GNU foundation hosted projects- it's dead. 

Instead of supporting these as separate service, why not just add support for the UI they're using: cgit.
I don't know much about it, and based on what you have so far it looks like there are some differences in how one can configure it, but maybe there's common ground. 

Thoughts?

kaushalmodi:
Thanks for reviewing the PR.

I added the support for emacs and org-mode git repos as I found myself having to get the absolute links with commits quite often. And I thought that emacs and org-mode devs might benefit from this.

I had to add them separately because the emacs git links have `/cgit/` in their paths and the org-mode git links have `/cgit.cgi/` in their paths.

Examples:
```
http://git.savannah.gnu.org/cgit/emacs.git/tree/admin/README?id=ab30bf5e8736f1414def53adae76a257fa482e53#n1
http://orgmode.org/cgit.cgi/org-mode.git/tree/README?id=1376123dea59e1aacdffef85af91273f7511e4dd#n1
```

Apart from the absence/presence of `.cgi` in the links, the link formats are identical. But it is not possible to add `.cgi` to emacs links or remove that from org-mode links.

As these links are for emacs and related org-mode git repos, I thought that git-link.el might be a good place for them. But if not, I can always update those alists in my personal emacs config.

sshaw:
> I had to add them separately because the emacs git links have /cgit/ in their paths and the org-mode git links have /cgit.cgi/ in their paths.

Indeed. 

It looks like [cgit defaults to `cgit.cgi`](https://git.zx2c4.com/cgit/tree/Makefile#n4) but of course this can be *anything*. cgit's site does not use `.cgi`.

I'm sure there's a way to work with this. Ideas? Supporting for cgit repos would be nice.

> As these links are for emacs and related org-mode git repos, I thought that git-link.el might be a good place for them. But if not, I can always update those alists in my personal emacs config.

Yes I would just add there, but again, I think it would be great to support cgit.

kaushalmodi:
Will think over the best approach, rebase and resend PR.

sshaw:
Would indeed be nice to have cgit support.

PR #31, opened 2016-04-15
Added git-link-latest-commit

calebmeyer:
Rationale: My code review workflow where I work requires that we provide
a link to a single commit for each change we make. That means you make
the change, commit it, push it up to your dev branch, and then add a
link of the commit to the code review.

I couldn't find anywhere in magit that automatically displays the hash
of the latest commit. This function allows me to simply get a link to
the latest commit copied to my kill-ring (and therefore my system
clipboard).

Implementation: I tried to follow exactly the conventions already
established. The only change from git-link-commit is that we use the
`(git-link-last-commit)` to get the commit hash.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ kill ring.
 
 ## Usage
 
-`M-x git-link` and `M-x git-link-commit`
+`M-x git-link`, `M-x git-link-commit`, and `M-x git-link-latest-commit`
 
 With a prefix argument prompt for the remote's name. Defaults to `"origin"`.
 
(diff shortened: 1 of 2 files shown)


sshaw:
Hi @calebmeyer, thanks for the pull request and sorry it to so long to reply. 

Can't one already achieve what you've described but setting [the `git-link-use-commit` option](https://github.com/sshaw/git-link#git-link-use-commit) to `t`?

[Support for Git Time Machine](https://github.com/sshaw/git-link#git-timemachine) has also been added. 

Let me know if these suffice.

calebmeyer:
Yeah, those will work. Thanks!

Issue #32, opened 2016-07-28
git link breaks on repositories accessed through custom port


kbrooks:
At my company we host our git repositories on gitlab and access them through port 2222.
For example, the url config on one of our repositories is:
	url = ssh://git@gitlab.imshealth.com:2222/tools/ImsHealth.Automation.git

The function git-link--remote-host fails when there is a port, simply returning "2222".

https://github.com/sshaw/git-link/blob/master/git-link.el#L163

It should return "gitlab.imshealth.com".

kbrooks:
I've started working on a fix for this issue

sshaw:
Hey, actually I forgot all the URL styles that were supported (I really need to add some tests!) so that didn't work either 😿 but [checkout this branch](https://github.com/sshaw/git-link/tree/url-with-port) and let me if that fixes it.

Thanks.

kbrooks:
That works for me

sshaw:
Fixed in `v0.4.2`.

PR #33, opened 2016-07-28
Fixing git-link-remote-regex to work with services on nonstandard ports

kbrooks:
https://github.com/sshaw/git-link/issues/32

Switched over to the rx macro because it's easier to read

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -1,7 +1,7 @@
 ;;; git-link.el --- Get the GitHub/Bitbucket/GitLab URL for a buffer location
 
 ;; Author: Skye Shaw <skye.shaw@gmail.com>
-;; Version: 0.4.1 (unreleased)
+;; Version: 0.4.1
 ;; Keywords: git
 ;; URL: http://github.com/sshaw/git-link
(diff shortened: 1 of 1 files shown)


kbrooks:
whoops this doesn't work

kbrooks:
I'm going to add a way for users to specify their own entries to git-link-remote-alist and git-link-commit-remote-alist.

git-link.el
```diff
@@ -105,7 +105,18 @@
...
+   (submatch
+    (one-or-more (in "-" "." word digit)))
+   (any ":" "/")
+   (submatch
+    (one-or-more
+     (one-or-more (not (in "/")))
+     "/"
+     (+? (not (in "/")))))
+   (? ".git")
+   eol))
```
sshaw: Hi, thanks for this but let's use `"\\([-.[:word:]]+\\(?::[0-9]+\\)?\\)/\\([^/]+/[^/]+?\\)\\(?:\\.git\\)?$"` instead. No need to bring in `rx`.

PR #34, opened 2016-08-06
Add savannah.gnu.org (cgit) support

itorres:
The main difference is the modification in [`git-link-remote-regex`](https://github.com/itorresgit-link/blob/add-savannah-support/git-link.el#L110) since cgit instances don't necessarily have an username part and even in savannah.gnu.org some projects live in the git root (for example: http://git.savannah.gnu.org/cgit/guix.git/)

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -90,22 +90,24 @@
   "If non-nil use the latest commit's hash in the link instead of the branch name.")
 
 (defvar git-link-remote-alist
-  '(("github.com"    git-link-github)
-    ("bitbucket.org" git-link-bitbucket)
-    ("gitorious.org" git-link-gitorious)
-    ("gitlab.com"    git-link-gitlab))
+  '(("github.com"           git-link-github)
+    ("bitbucket.org"        git-link-bitbucket)
+    ("gitorious.org"        git-link-gitorious)
+    ("gitlab.com"           git-link-gitlab)
+    ("git.savannah.gnu.org" git-link-cgit))
   "Maps remote hostnames to a function capable of creating the appropriate file URL")
 
 (defvar git-link-commit-remote-alist
-  '(("github.com"    git-link-commit-github)
-    ("bitbucket.org" git-link-commit-bitbucket)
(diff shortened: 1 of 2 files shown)


sshaw:
Hi @itorres, thanks for this commit. 

Support for Savanah has [been discussed before](https://github.com/sshaw/git-link/pull/29#issuecomment-190507750) and my feelings can be found [here](https://github.com/sshaw/git-link/pull/29#issuecomment-190507750).

Support for cgit would be great (🎆 🎉 ❗) but there's [a small quirk](https://github.com/sshaw/git-link/pull/29#issuecomment-195653908). 
How do you propose dealing with this? 

I think I created a branch locally to simply service creation and better deal with different url formats by allowing one to do things like this:

```elisp
  '("foo.com" "http://{hostname}/cgit.cgi/{dirname}}/commit?id={{commit}}"))
  '("example.com" old-style-funtion))
```

But it's not done. 

Also can you rebase against the [`url-with-port` branch](https://github.com/sshaw/git-link/tree/url-with-port) and fix your regex additions to match it. That branch resolves #32. 

Thanks!

git-link.el
```diff
@@ -90,22 +90,24 @@
...
 (defvar git-link-commit-remote-alist
-  '(("github.com"    git-link-commit-github)
-    ("bitbucket.org" git-link-commit-bitbucket)
-    ("gitorious.org" git-link-commit-gitorious)
-    ("gitlab.com"    git-link-commit-github))
+  '(("github.com"           git-link-commit-github)
+    ("bitbucket.org"        git-link-commit-bitbucket)
+    ("gitorious.org"        git-link-commit-gitorious)
+    ("gitlab.com"           git-link-commit-github)
+    ("git.savannah.gnu.org" git-link-commit-cgit))
```
sshaw: Depending on [how one cloned the repo](http://git.savannah.gnu.org/cgit/zdl.git/), the remote can also point to git.sv.gnu.org.

sshaw:
Would be great to have cgit support. Please feel free to reopen if the outstanding issues/questions can be resolved.

Issue #37, opened 2017-01-28
Return link from main commands


dougharris:
There's currently no way to get the link for use in other elisp functions. It'd be nice if the commands returned the link in addition to writing it to the Messages buffer.

I've got a one link fix and I'll submit a pull request.

PR #38, opened 2017-01-28
Return link

dougharris:
https://github.com/sshaw/git-link/issues/37

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -241,7 +241,9 @@
   (message (replace-regexp-in-string "%" "%%" link t t))
   (setq deactivate-mark t)
   (when git-link-open-in-browser
-    (browse-url link)))
+    (browse-url link))
+  link
+  )
 
 (defun git-link-gitlab (hostname dirname filename branch commit start end)
   (format "https://%s/%s/blob/%s/%s#%s"


sshaw:
Hi, thanks for this. It's been something I've been meaning to do for a while. 

One question: don't you think it would be better to move this to a non-interactive function that doesn't touch the kill ring or open the link? 

Something like:

```el
(defun git-link-url (remote &optional start end)
  ;; build url
)

(defun git-link (remote start end)
  ;; interactive stuff
  (git-link--new (git-link-url remote start end)))
```

truthdoug:
Yes - non-interactive would be better. The other thing I was thinking about was a shortcut that wouldn't require the `remote` param because most uses have only a single remote.

sshaw:
The reason I included remote is so one can use it without having to depend on a buffer's contents. 

I guess on could always default to the current buffer, but in general I think it's good for functions to not assume anything about global state.

sshaw:
Also, what's your use case for getting the URL?

dougharris:
I have a hand made, half-assed function that opens the commit but it relies on a manually maintained map of directories to remote repo URLs. I found git-link and it did much of the hard stuff that I wasn't doing.

So, I'm basically looking for the `git-link-commit` functionality but leaving without setting a variable that always opens the link. There are times where I want the URL without opening (e.g. for sharing with colleagues).

futuro:
Hey y'all, I'd love to help push this across the finish line! My use case is I'm writing out org-mode capture templates and I want to be able to programmatically call git-link so that I share my exports with others in a meaningful way.

Let me know what I can do to help!

sshaw:
Closing to to lack of activity and #69

PR #39, opened 2017-02-16
Fix string quoting

kaushalmodi:
This matches the change made to text formatting and quoting styles in emacs
25.1.

From (elisp) Displaying Messages:

> The ‘text-quoting-style’ variable controls what quotes are
> generated; *Note Keys in Documentation::.  A call using a format
> like "Missing `%s'" with grave accents and apostrophes typically
> generates a message like "Missing ‘foo’" with matching curved
> quotes.  In contrast, a call using a format like "Missing '%s'"
> with only apostrophes typically generates a message like "Missing
> ’foo’" with only closing curved quotes, an unusual style in
> English.

Now,

    (message "Reverting `%s'..." "foo.el")

prints as:

    Reverting ‘foo.el’...

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -328,7 +328,7 @@ Defaults to \"origin\"."
     (cond ((null filename)
 	   (message "Not in a git repository with a working tree"))
 	  ((null remote-host)
-	   (message "Remote '%s' is unknown or contains an unsupported URL" remote))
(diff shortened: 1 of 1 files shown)


sshaw:
Interesting v25 feature. Thanks!

PR #40, opened 2017-02-17
Allow subdomain git remotes

kaushalmodi:
**UPDATE (2017/02/18)**

The `git-link` and `git-commit-link` functions will now work even if the
remote-host is not exactly "gitlab.com", but something like "gitlab.COMPANY.com"
or "gitlab.COMPANY.org" or "gitlab.COMPANY.co.in".

This commits adds the use of `cl-loop` macro. If using emacs 24.x or newer,
this is in-built in emacs in `cl-macs` library (which is required by the `cl-lib`
library).

But for older emacs versions, the `cl-lib` library will be auto-installed from
GNU Elpa with the help of the Package-Requires line added in this commit:

    ;; Package-Requires: ((cl-lib "0.6.1"))

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,26 +68,39 @@ Name of the remote branch to link to.
 
 ### Git Timemachine
 
-If [`git-timemachine-mode`](https://github.com/pidu/git-timemachine) is active `git-link` generates a URL for the version of the file being visited.
+If [`git-timemachine-mode`](https://github.com/pidu/git-timemachine)
+is active `git-link` generates a URL for the version of the file being
+visited.
 
 ### Building Links and Adding Services
 
-`git-link-remote-alist` and `git-link-commit-remote-alist` map remotes'
-hostnames to a function capable of creating a URL on that host. To add (or
-modify) how URLs are created for a given host add the appropriate function
-objects to these lists.
+`git-link-remote-alist` is an alist containing `(REGEXP FUNCTION)`
+elements. The FUNCTION creates URLs for file on remote host names that
+match the REGEXP. To add (or modify) how URLs are created for a given
+host, add appropriate elements to this list.
 
-If you use a self-hosted version of one of the supported services, you
-can configure the link function alists for the hostname at which that
-service is hosted.  For example, for a GitHub Enterprise instance at
-`github.example.com`, you could add the following to your `.emacs` file.
+As an example, one of the default elements in this alist is
+`("gitlab" git-link-gitlab)`. So the `git-link-gitlab` function
+will be used to create URLs to files in remotes that match the
+*regexp* `"gitlab"`.  That would cover common Gitlab host URLs like
+*"gitlab.com"*, *"gitlab.example.com"* and *"gitlab.example.org"*.
+
+`git-link-commit-remote-alist` is also an alist containing `(REGEXP
+FUNCTION)` elements. Here, the FUNCTION creates URLs to the commit
+pages, for remote hosts matching REGEXP.
+
+If you use a self-hosted version of one of the supported services, but
+your remote URL does match with the defaults, you can configure these
+link function alists.  For example, for a GitHub Enterprise instance
+at `gh.example.com`, you could add the following to your `.emacs`
+file:
 
     (eval-after-load 'git-link
       '(progn
         (add-to-list 'git-link-remote-alist
-          '("github.example.com" git-link-github))
+          '("gh\\.example\\.com" git-link-github))
         (add-to-list 'git-link-commit-remote-alist
-          '("github.example.com" git-link-commit-github))))
+          '("gh\\.example\\.com" git-link-commit-github))))
 
 The `git-link` signature is:
 
--- a/git-link.el
+++ b/git-link.el
@@ -5,6 +5,7 @@
 ;; Version: 0.5.0 (unreleased)
 ;; Keywords: git, vc, github, bitbucket, gitlab, convenience
 ;; URL: http://github.com/sshaw/git-link
+;; Package-Requires: ((cl-lib "0.6.1"))
 
 ;; This file is NOT part of GNU Emacs.
 
@@ -88,6 +89,7 @@
 
 ;;; Code:
 
+(require 'cl-lib)
 (require 'thingatpt)
 (require 'url-util)
 
@@ -107,18 +109,28 @@
   "If non-nil use the latest commit's hash in the link instead of the branch name.")
 
 (defvar git-link-remote-alist
-  '(("github.com"    git-link-github)
-    ("bitbucket.org" git-link-bitbucket)
-    ("gitorious.org" git-link-gitorious)
-    ("gitlab.com"    git-link-gitlab))
-  "Maps remote hostnames to a function capable of creating the appropriate file URL")
+  '(("github" git-link-github)
+    ("bitbucket" git-link-bitbucket)
+    ("gitorious" git-link-gitorious)
+    ("gitlab" git-link-gitlab))
+  "Alist containing (REGEXP FUNCTION) where REGEXP is used to
+match the remote host URL and FUNCTION is capable of creating the
+appropriate file URL.
+
+As an example, \"gitlab\" will match with both \"gitlab.com\" and
+\"gitlab.example.com\".")
 
 (defvar git-link-commit-remote-alist
-  '(("github.com"    git-link-commit-github)
-    ("bitbucket.org" git-link-commit-bitbucket)
-    ("gitorious.org" git-link-commit-gitorious)
-    ("gitlab.com"    git-link-commit-github))
-  "Maps remote hostnames to a function capable of creating the appropriate commit URL")
+  '(("github" git-link-commit-github)
+    ("bitbucket" git-link-commit-bitbucket)
+    ("gitorious" git-link-commit-gitorious)
+    ("gitlab" git-link-commit-github))
+  "Alist containing (REGEXP FUNCTION) where REGEXP is used to
+match the remote host URL and FUNCTION is capable of creating the
+appropriate commit URL.
+
+As an example, \"gitlab\" will match with both \"gitlab.com\" and
+\"gitlab.example.com\".")
 
 ;; Matches traditional URL and scp style:
 ;; https://example.com/ruby/ruby.git
@@ -180,6 +192,23 @@
     (if (or (null remote) (string= remote "."))
 	"origin"
       remote)))
+
+(defun git-link--handler (alist str)
+  "For an ALIST whose `car' (a regexp) matches STR, return cadr.
+
+The ALIST consists of (REGEXP FN) list elements.
+Valid ALISTs are `git-link-commit-remote-alist',`git-link-commit-alist'.
+
+For the first ALIST element whose REGEXP matches with STR, FN is
+returned.
+
+Return nil,
+- if STR does not match with REGEXP in any of the elements of ALIST, or
+- if STR is not a string"
+  (when (stringp str)
+    (cadr (cl-find-if (lambda (lst)
+                        (string-match-p (car lst) str))
+                      alist))))
 
 (defun git-link--relative-filename ()
   (let* ((filename (buffer-file-name))
@@ -325,11 +354,10 @@ Defaults to \"origin\"."
                       (region (git-link--get-region)))
                  (list remote (car region) (cadr region))))
   (let* ((remote-host (git-link--remote-host remote))
-	 (filename    (git-link--relative-filename))
-	 (branch      (git-link--branch))
-	 (commit      (git-link--commit))
-	 (handler     (cadr (assoc remote-host git-link-remote-alist))))
-
+	 (filename (git-link--relative-filename))
+	 (branch (git-link--branch))
+	 (commit (git-link--commit))
+	 (handler (git-link--handler git-link-remote-alist remote-host)))
     (cond ((null filename)
 	   (message "Not in a git repository with a working tree"))
 	  ((null remote-host)
@@ -360,8 +388,8 @@ Defaults to \"origin\"."
 
   (interactive (list (git-link--select-remote)))
   (let* ((remote-host (git-link--remote-host remote))
-	 (commit      (word-at-point))
-	 (handler     (cadr (assoc remote-host git-link-commit-remote-alist))))
+	 (commit (word-at-point))
+	 (handler (git-link--handler git-link-commit-remote-alist remote-host)))
     (cond ((null remote-host)
 	   (message "Remote `%s' is unknown or contains an unsupported URL" remote))
 	  ((not (string-match "[a-z0-9]\\{7,40\\}" (or commit "")))
--- a/git-link.el
+++ b/git-link.el
@@ -1,4 +1,4 @@
-;;; git-link.el --- Get the GitHub/Bitbucket/GitLab URL for a buffer location
+;;; git-link.el --- Get the GitHub/Bitbucket/GitLab URL for a buffer location -*- lexical-binding: t -*-
 
 ;; Copyright (C) 2013-2017 Skye Shaw and others
 ;; Author: Skye Shaw <skye.shaw@gmail.com>
@@ -27,11 +27,11 @@
 ;;; Commentary:
 
 ;; Create URLs for files and commits in GitHub/Bitbucket/GitLab/...
-;; repositories. `git-link' returns the URL for the current buffer's file
-;; location at the current line number or active region. `git-link-commit'
-;; returns the URL for a commit. URLs are added to the kill ring.
+;; repositories.  `git-link' returns the URL for the current buffer's file
+;; location at the current line number or active region.  `git-link-commit'
+;; returns the URL for a commit.  URLs are added to the kill ring.
 ;;
-;; With a prefix argument prompt for the remote's name. Defaults to "origin".
+;; With a prefix argument prompt for the remote's name.  Defaults to "origin".
 
 ;;; Change Log:
 
@@ -113,9 +113,10 @@
     ("bitbucket" git-link-bitbucket)
     ("gitorious" git-link-gitorious)
     ("gitlab" git-link-gitlab))
-  "Alist containing (REGEXP FUNCTION) where REGEXP is used to
-match the remote host URL and FUNCTION is capable of creating the
-appropriate file URL.
+  "Alist of host names and functions creating file links for those.
+Each element looks like (REGEXP FUNCTION) where REGEXP is used to
+match the remote's host name and FUNCTION is used to generate a link
+to the file on remote host.
 
 As an example, \"gitlab\" will match with both \"gitlab.com\" and
 \"gitlab.example.com\".")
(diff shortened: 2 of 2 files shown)


sshaw:
Hi. Thanks but this is [already possible](https://github.com/sshaw/git-link#building-links-and-adding-services).

git-link.el
```diff
@@ -103,18 +103,25 @@
...
 (defvar git-link-remote-alist
-  '(("github.com"    git-link-github)
-    ("bitbucket.org" git-link-bitbucket)
-    ("gitorious.org" git-link-gitorious)
-    ("gitlab.com"    git-link-gitlab))
-  "Maps remote hostnames to a function capable of creating the appropriate file URL")
+  '(("github\\.\\(.+\\.\\)*com" git-link-github)
+    ("bitbucket\\.\\(.+\\.\\)*org" git-link-bitbucket)
+    ("gitorious\\.\\(.+\\.\\)*org" git-link-gitorious)
+    ("gitlab\\.\\(.+\\.\\)*com" git-link-gitlab))
```
sshaw: What if the company's top-level domain is not `.com`?

kaushalmodi:
You're right. Will close this PR.

git-link.el
kaushalmodi: I got lazy and made it fit what worked for me.

kaushalmodi:
I still like the regexp approach more. What if that regexp is made a defcustom?

kaushalmodi:
.. Or, instead of a regexp defcustom, the default regexp for Gitlab can be just `"gitlab"`. I believe that will work for most people.

 If that doesn't work, then they can just the regexp that works for them using the `add-to-list` method.

kaushalmodi:
I have force-pushed an update to my origin branch. Please review again.

- Made the regexp generic: `"gitlab"`, `"github"`, etc.
- Added require for `cl-lib`.

sshaw:
> .. Or, instead of a regexp defcustom, the default regexp for Gitlab can be just "gitlab". I believe that will work for most people.

I mean I do kinda like this idea, but it presents problems if say if my repository is at `gitlab.com/sshaw/github-client`. It will create a link to `http://github.com/sshaw/github-client/blob/master/foo.whatever`

sshaw:
Wait, actually no it wont. Almost forgot how this thing works 😄. Let me have an in depth look.

kaushalmodi:
Yeah, that regexp applies to only the remote base "github

kaushalmodi:
Ah! On phone.. mis-clicked on Close. 

Continuing.. remote base "github.com".

kaushalmodi:
As a side, I just happened to notice the TODO: * Consolidate `git-link-*-alist's*.

I was just thinking why we had 2 alists.. how about elements of type *(REMOTE-LINK-REGEXP . (FN-LINK FN-COMMIT-LINK))*?

If you are willing, I can work on that PR.

kaushalmodi:
Made another forced push with README update.

git-link.el
```diff
@@ -323,8 +335,9 @@ Defaults to \"origin\"."
 	 (filename    (git-link--relative-filename))
 	 (branch      (git-link--branch))
 	 (commit      (git-link--commit))
-	 (handler     (cadr (assoc remote-host git-link-remote-alist))))
-
+	 (handler     (cl-loop for (key . value) in git-link-remote-alist
+                               when (string-match-p key remote-host)
```
sshaw: `remote-host` can be `nil` if the repository has no remotes.

git-link.el
```diff
@@ -356,7 +369,9 @@ Defaults to \"origin\"."
   (interactive (list (git-link--select-remote)))
   (let* ((remote-host (git-link--remote-host remote))
 	 (commit      (word-at-point))
-	 (handler     (cadr (assoc remote-host git-link-commit-remote-alist))))
+	 (handler     (cl-loop for (key . value) in git-link-commit-remote-alist
+                               when (string-match-p key remote-host)
+                               finally return (car value))))
```
sshaw: Do you think these `cl-loop` calls should be put into a function?

git-link.el
```diff
@@ -103,18 +105,28 @@
...
-  '(("github.com"    git-link-github)
-    ("bitbucket.org" git-link-bitbucket)
-    ("gitorious.org" git-link-gitorious)
-    ("gitlab.com"    git-link-gitlab))
-  "Maps remote hostnames to a function capable of creating the appropriate file URL")
+  '(("github" git-link-github)
+    ("bitbucket" git-link-bitbucket)
+    ("gitorious" git-link-gitorious)
+    ("gitlab" git-link-gitlab))
+  "Alist containing (REMOTE-HOST-REGEXP FUNCTION) where REMOTE-HOST-REGEXP is
```
sshaw: I think just `REGEXP` is fine since you go on to say:

> ... a regexp to match the remote host URL

sshaw:
> But for older emacs versions, the cl-lib library will be auto-installed...
> with the help of the Package-Requires line added in this commit:
> ;; Package-Requires: ((cl-lib "0.6.1"))

Why this version. I would think 0.5. 

> I was just thinking why we had 2 alists..
> how about elements of type (REMOTE-LINK-REGEXP . (FN-LINK FN-COMMIT-LINK))?

Not sure on this, though offhand I don't know what the other options are.

What about `git-link-homepage-XXXX`, would that just get tacked on to the value list?
(Now there's just 1 function for this, but it's possible this may need to be version specific.)

Regardless I'd keep both alists around with the new scheme (whatever that would be) for sometime to allow folks to migrate. 

> If you are willing, I can work on that PR.

I think adding tests are more important than alist consolidation 😸

git-link.el
kaushalmodi: Sure. Will do. Will also put a check for remote "nil-ness" in there.

kaushalmodi:
0.6.1 is the latest cl-lib version available on GNU Elpa. Is there a reason to use version 0.5?

Learning to use ert is on my list.. some day. I am working on few packages, so will learn to write tests eventually.

git-link.el
sshaw: Also, this always returns the last element in the alist: `finally return (car value)`.
We need to break (`return`) if a match is found. `cl-find` can work too.

sshaw:
> Is there a reason to use version 0.5?

Compatibility with versions <= 23.3 (I think) for free. Not sure of any downsides. It's what I've used elsewhere.

git-link.el
kaushalmodi: OK; will make that change.

git-link.el
kaushalmodi: Will do more testing with this before my next update. It was my first time using `cl-loop`. Will have a look at `cl-find` too (never used it).

kaushalmodi:
> Compatibility with versions <= 23.3 (I think) for free. Not sure of any downsides. It's what I've used elsewhere.

The commentary in [cl-lib 0.6.1][1] also says:

> ;; This is a forward compatibility package, which provides (a subset of) the
> ;; features of the cl-lib package introduced in Emacs-24.3, for use on
> ;; previous emacsen (it should work on Emacs≥21 as well as XEmacs).

I would say that it is safe to use 0.6.1.

The 0.6 and 0.6.1 updates happened very recently in Jan 2017. So that's why it's not yet out in the wild.

> > how about elements of type (REMOTE-LINK-REGEXP . (FN-LINK FN-COMMIT-LINK))?
> Not sure on this, though offhand I don't know what the other options are.

Does it mean it's OK to go ahead with that? :)

> What about git-link-homepage-XXXX, would that just get tacked on to the value list?
> (Now there's just 1 function for this, but it's possible this may need to be version specific.)

Yes, why not. We can put that same fn name in all lists for now.

> Regardless I'd keep both alists around with the new scheme (whatever that would be) for sometime to allow folks to migrate.

That's understood.

[1]: https://elpa.gnu.org/packages/cl-lib-0.6.1.el

kaushalmodi:
Made another forced update on this PR branch.

Now using `cl-find-if` instead of `cl-loop`. 

`cl-seq` will be autoloaded by `cl-find-if`. So doesn't need any explicit requires.

Here is my poor man test for this change:

```el
(let ((hosts '("gitlab.com"
               "gitlab.example.com"
               nil
               "gh.com"
               "bitbucket.com"
               "github.com"
               "github.example.com"
               "bitbucket.com"
               "gh.example.com"
               "gitorious.com"))
      (alist '(("github" git-link-github)
               ("bitbucket" git-link-bitbucket)
               ("gitorious" git-link-gitorious)
               ("gitlab" git-link-gitlab)
               ("gh\\.example" git-link-gitlab)))
      (n 1)
      host)
  (dolist (host hosts)
    (message "host = %20S  returned fn = %S" host
             (when (stringp host)
               (nth n (cl-find-if (lambda (lst)
                                    (string-match-p (car lst) host))
                                  alist)))))
  nil)
```

git-link.el
kaushalmodi: Now fixed.

git-link.el
kaushalmodi: Now fixed, using `cl-find-if`, thanks!

git-link.el
kaushalmodi: > I think just REGEXP is fine 

Done.

git-link.el
```diff
@@ -176,6 +186,22 @@
...
+The ALIST consists of (REGEXP FN) list elements.
+Valid ALISTs are `git-link-commit-remote-alist',`git-link-commit-alist'.
+
+Nth element of ALIST is returned at the first match of STR with REGEXP.
+
+Return nil,
+- if STR does not match with REGEXP in any of the elements of ALIST, or
+- if STR is not a string "
+  (when (stringp remote-host)
+    (nth n (cl-find-if (lambda (lst)
```
sshaw: Forgot to `(require 'cl-lib)` and add the `Package-Requires`

sshaw: > cl-seq will be autoloaded by cl-find-if. So doesn't need any explicit requires.

I don't follow.

kaushalmodi: Yeah.. my mistake. I somehow thought that cl-lib won't be required. Adding it back.

sshaw:
>>> how about elements of type (REMOTE-LINK-REGEXP . (FN-LINK FN-COMMIT-LINK))?
>> Not sure on this, though offhand I don't know what the other options are.
> Does it mean it's OK to go ahead with that? :)

I'd say hold-off for now. Need to think about it a bit more.

git-link.el
```diff
@@ -103,18 +103,28 @@
...
+  '(("github" git-link-commit-github)
+    ("bitbucket" git-link-commit-bitbucket)
+    ("gitorious" git-link-commit-gitorious)
+    ("gitlab" git-link-commit-github))
+  "Alist containing (REGEXP FUNCTION) where REGEXP is used to
+match the remote host URL and FUNCTION is capable of creating the
+appropriate commit URL.
+
+As an example, \"gitlab\" will match with both \"gitlab.com\" and
+\"gitlab.COMPANY.com\".")
```
sshaw: Since doc strings use capitals for arguments should "COMPANY"  really be capitalized here?

git-link.el
sshaw: Though now that `cl-find-if` is being used is this function necessary? 😬

Also `nth` is not needed since we'll hold off on alist consolidation.

git-link.el
kaushalmodi: Fixed in my latest push.. *COMPANY* -> *example*

git-link.el
kaushalmodi: Fixed in latest push.

README.md
```diff
@@ -68,26 +68,39 @@ Name of the remote branch to link to.
...
+As an example, one of the default elements in this alist is
+`("gitlab" git-link-gitlab)`. So the `git-link-gitlab` function
+will be used to create URLs to files in remotes that match the
+*regexp* `"gitlab"`.  That would cover common Gitlab host URLs like
+*"gitlab.com"*, *"gitlab.example.com"* and *"gitlab.example.org"*.
+
+`git-link-commit-remote-alist` is also an alist containing `(REGEXP
+FUNCTION)` elements. Here, the FUNCTION creates URLs to the commit
+pages, for remote hosts matching REGEXP.
+
```
sshaw: I have some minor nitpicks here but I don't want to beat you over the head too much about this after all your work. I will update later.

git-link.el
kaushalmodi: > Though now that cl-find-if is being used is this function necessary? 

I just put it in for the `stringp` check. Also allows a good place to explain what it does in the fn's docstring.

> Also nth is not needed since we'll hold off on alist consolidation.

Would it be alright to keep that in.. just in case?

README.md
kaushalmodi: No worries. Learning here.. how can I improve that?

kaushalmodi:
I pushed the changes as you commented on the commits yesterday. Let me know if you are waiting for something from me.

git-link.el
```diff
@@ -176,6 +188,22 @@
...
+Valid ALISTs are `git-link-commit-remote-alist',`git-link-commit-alist'.
+
+Nth element of ALIST is returned at the first match of STR with REGEXP.
+
+Return nil,
+- if STR does not match with REGEXP in any of the elements of ALIST, or
+- if STR is not a string "
+  (when (stringp remote-host)
+    (nth n (cl-find-if (lambda (lst)
+                         (string-match-p (car lst) remote-host))
```
sshaw: `remote-host` is being resolved dynamically. Argument is `alist`.

git-link.el
sshaw: Who knows if there will ever be a need for `nth N`. Right now just nothing but cause one to scratch their head and wonder why it's there 😄. It's simple enough to add when/if it's needed.

git-link.el
```diff
@@ -103,18 +105,28 @@
...
-    ("bitbucket.org" git-link-commit-bitbucket)
-    ("gitorious.org" git-link-commit-gitorious)
-    ("gitlab.com"    git-link-commit-github))
-  "Maps remote hostnames to a function capable of creating the appropriate commit URL")
+  '(("github" git-link-commit-github)
+    ("bitbucket" git-link-commit-bitbucket)
+    ("gitorious" git-link-commit-gitorious)
+    ("gitlab" git-link-commit-github))
+  "Alist containing (REGEXP FUNCTION) where REGEXP is used to
+match the remote host URL and FUNCTION is capable of creating the
```
sshaw: "the remote's hostname and FUNCTION..."

git-link.el
kaushalmodi: Correct. I couldn't though figure out how to pass it otherwise. The only arg lambda can accept is the alist element. Ideas?

git-link.el
```diff
@@ -103,18 +105,28 @@
...
-    ("bitbucket.org" git-link-bitbucket)
-    ("gitorious.org" git-link-gitorious)
-    ("gitlab.com"    git-link-gitlab))
-  "Maps remote hostnames to a function capable of creating the appropriate file URL")
+  '(("github" git-link-github)
+    ("bitbucket" git-link-bitbucket)
+    ("gitorious" git-link-gitorious)
+    ("gitlab" git-link-gitlab))
+  "Alist containing (REGEXP FUNCTION) where REGEXP is used to
+match the remote host URL and FUNCTION is capable of creating the
```
sshaw: "the remote's hostname and FUNCTION..."

git-link.el
sshaw: It should work. Maybe you're thinking about returning a closure over the functions arguments?

sshaw: But also, maybe this is a sign that the function is not needed 😺

kaushalmodi: Ah, I realized the mistake. Need to use str instead of remote-host.

kaushalmodi:
Please review; I have made couple of doc-string and compilation warning fixes.

- Ran `checkdoc` (should have done that earlier).
- Most of the functions lack doc-strings; so `checkdoc` throws errors for those.
- Getting free variable warning for `git-timemachine-revision` var. I'll leave it up to you for the best way to fix that. I would put a `defvar-local` (but then that won't work for emacsen older than 24.x).

git-link.el
kaushalmodi: Done.

git-link.el
kaushalmodi: Done.

git-link.el
kaushalmodi: Fixed. Also enabled lexical-binding. Now compiles fine except for one free variable warning for `git-timemachine-revision`.

kaushalmodi:
Now compilation is squeaky clean even with lexical-binding enabled.

git-link.el
```diff
@@ -269,7 +299,7 @@
 	  dirname
 	  commit))
 
-(defun git-link-gitorious (hostname dirname filename branch commit start end)
+(defun git-link-gitorious (hostname dirname filename _branch commit start _end)
```
sshaw: Is `_unused_var` an emacs convention? Never seen this. Not a fan.

kaushalmodi: Yes, it's a convention, more than that it's the elisp compiler feature. On phone, but you should find this link in elisp manual, and in use at many places in emacs core.

git-link.el
```diff
@@ -320,11 +350,10 @@ Defaults to \"origin\"."
...
   (let* ((remote-host (git-link--remote-host remote))
-	 (filename    (git-link--relative-filename))
-	 (branch      (git-link--branch))
-	 (commit      (git-link--commit))
-	 (handler     (cadr (assoc remote-host git-link-remote-alist))))
-
+	 (filename (git-link--relative-filename))
+	 (branch (git-link--branch))
+	 (commit (git-link--commit))
+	 (handler (git-link--handler 1 git-link-remote-alist remote-host)))
```
sshaw: Still don't like having to pass `1` every time. We're coding for an abstraction that doesn't exist. Maybe it will, maybe it wont.

git-link.el
sshaw: Oh I see, the warning is emitted due to `lexical-binding: t`.  👍 

Just started trying to use that, actually. I'm a dinosaur on the Emacs front 👴.

git-link.el
kaushalmodi: Done!

git-link.el
kaushalmodi: No worries, thanks for accepting this change.

kaushalmodi:
I have rebased this PR to your master and replaced `nth 1` with `cadr`. Please review .

sshaw:
Hi, looks like the rebase is messed up. Take a look please.

kaushalmodi:
Sorry, how do I know that rebasing is wrong? I see this.. 3 commits based to the current upstream (sshaw/git-link) master.

image

sshaw:
Looks like you've merged things into your branch, then rebased against master, then merged some more stuff into your branch 😕.

A few solutions come to mind: 
1. Drop the merge commits (not shown in above screenshots, but see with `git log`) and then rebase. 

1. Create a patch file from a7123cdd4ae2c55cc59dd17c1c657ef6ca3c04f4, 79d6a590581e01137dbb6cc1e49e3b4d9fda1729, and 
b403c1c2e7fb408a00051c05b082986cb2227840 and just apply that to a new branch (though b403c1c and a7123 should be squashed into a singe commit IMHO)


I think the 2nd may be better (untested): 

```
git log -1 -p a7123cdd4ae2c55cc59dd17c1c657ef6ca3c04f4 > /tmp/p1.patch
git log -1 -p b403c1c2e7fb408a00051c05b082986cb2227840 > /tmp/p1.patch
git log -1 -p 79d6a590581e01137dbb6cc1e49e3b4d9fda1729 >> /tmp/p2.patch
git checkout -b new-branch  master
git apply /tmp/p1.patch
git commit -a
git apply /tmp/p2.patch
git commit -a 
```

sshaw:
Hurrmmm actually options 2 seems like it may be more work.

kaushalmodi:
Please check now. I finally understood my mistake.. I was doing merges instead of rebases. Magit made it a breeze to fix this.

sshaw:
Great, thanks! 😅

PR #41, opened 2017-02-17
Relax the git-link-remote-regex to allow more valid links

kaushalmodi:
This changes a minor portion of `git-link-remote-regex` as follows:

```
     Before: "\\([^/]+/[^/]+?\\)"
     After:  "\\([^/]+\\(?:/[^/]+?\\)*\\)".
```

Earlier this regexp did not match "http://orgmode@orgmode.org/org-mode.git".

Now it does; without breaking the match for links containing 2 forward-slashes
as usual too -- "https://git.savannah.gnu.org/r/emacs.git"

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -120,10 +120,11 @@
 ;; https://example.com/ruby/ruby.git
 ;; git@example.org:sshaw_/customer_gender.git
 ;; git@example.com:22/foo/bar.git
+;; http://orgmode@orgmode.org/org-mode.git
 ;;
 ;; Wont work for git remotes that aren't services.
 ;; Consider using url-generic-parse-url, but that requires a URL with a scheme
-(defconst git-link-remote-regex "\\([-.[:word:]]+\\)\\(?:/\\|:[0-9]*/?\\)\\([^/]+/[^/]+?\\)\\(?:\\.git\\)?$")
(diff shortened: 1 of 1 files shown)


git-link.el
```diff
@@ -120,10 +120,11 @@
 ;; https://example.com/ruby/ruby.git
 ;; git@example.org:sshaw_/customer_gender.git
 ;; git@example.com:22/foo/bar.git
+;; http://orgmode@orgmode.org/org-mode.git
 ;;
 ;; Wont work for git remotes that aren't services.
 ;; Consider using url-generic-parse-url, but that requires a URL with a scheme
-(defconst git-link-remote-regex "\\([-.[:word:]]+\\)\\(?:/\\|:[0-9]*/?\\)\\([^/]+/[^/]+?\\)\\(?:\\.git\\)?$")
+(defconst git-link-remote-regex "\\([-.[:word:]]+\\)\\(?:/\\|:[0-9]*/?\\)\\([^/]+\\(/[^/]+?\\)*\\)\\(?:\\.git\\)?$")
```
sshaw: Any reason you're capturing the this path segment? I.e., one can now do `(match-string 3 url)`

kaushalmodi: Sorry, I had no reason to do so. I have force pushed to this PR branch with the fix .. added `?:`.

sshaw:
Thanks!

PR #42, opened 2017-02-22
Fix "reference to free variable" warning: `git-timemachine-revision`

kaushalmodi:


git-link.el
```diff
@@ -90,6 +90,9 @@
 (require 'thingatpt)
 (require 'url-util)
 
+(eval-when-compile
+  (defvar git-timemachine-revision))    ;silence reference to free variable warning
```
sshaw: I think`defvar` enough, no?

kaushalmodi: That leads to minor optimization as git-link does not need to evaluate that each time it is loaded. All we need is to tell the compiler that variable is an intended global declaration. https://www.gnu.org/software/emacs/manual/html_node/elisp/Eval-During-Compile.html

sshaw: > That leads to minor optimization...

Minor or micro 😸?

kaushalmodi:
Thanks :)

PR #43, opened 2017-02-28
Fix regexp for commit hash; support uppercase hash too

kaushalmodi:
Commit hash is a hexadecimal code -- 0..9, a..f

Also support commit hash in upper case.

Use `string-match-p` instead of `string-match` as the regexp
comparison is done just for boolean check.

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -394,7 +394,7 @@ Defaults to \"origin\"."
 	 (handler (git-link--handler git-link-commit-remote-alist remote-host)))
     (cond ((null remote-host)
 	   (message "Remote `%s' is unknown or contains an unsupported URL" remote))
-	  ((not (string-match "[a-z0-9]\\{7,40\\}" (or commit "")))
+	  ((not (string-match-p "[a-fA-F0-9]\\{7,40\\}" (or commit "")))
 	   (message "Point is not on a commit hash"))
 	  ((not (functionp handler))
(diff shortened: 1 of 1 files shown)


git-link.el
```diff
@@ -360,11 +360,11 @@ Defaults to \"origin\"."
 
   (interactive (list (git-link--select-remote)))
   (let* ((remote-host (git-link--remote-host remote))
-	 (commit      (word-at-point))
+	 (commit      (downcase (word-at-point)))
```
sshaw: While I do program paranoid under what circumstances does git return uppercase letters in SHAs?

When `word-at-pont` returns `nil` `downcase` will signal an error.

sshaw: > While I do program paranoid under what circumstances does git return uppercase letters in SHAs?

Better asked as: when will one encounter an uppercase letter in a SHA?

May as well just add `A-Z` to regex.

kaushalmodi: Ah, keep on missing the non-string case. Yeah, I was thinking of adding `A-F` to the regexp instead, will do that now. 

>  will one encounter an uppercase letter in a SHA?

I haven't encountered that. But I think it should be supported as uppercase hex is legal too.

kaushalmodi:
I have force pushed the fix to this branch.

sshaw:
Thanks!

PR #48, opened 2017-12-12
If no branch can be identified, use git rev-parse HEAD

jwiegley:
This makes it possible to get a git-link for detached HEADs, which commonly
happens with submodules.


sshaw:
Hi, thanks for the PR. 

A similar issue came up in #46. The concern I had is what to do if one has set `git-link-use-commit` to `nil`. In this case we'd still use the commit. The gist of the conversation can be found starting [here](https://github.com/sshaw/git-link/pull/46#issuecomment-331640957).

sshaw:
Closing. Will address via #46 as part of future 1.0 release.

PR #53, opened 2018-05-25
Add variable to allow the recognition of vc names

### LINKED ISSUE #52: Infer revision used when visiting buffer created by vc-find-revision
I like to use `vc-revision-other-window` to checkout old versions of a file in emacs. It would be awesome if git-link understood what was happening in this case and would like to the appropriate revision/filename when I use it from one of these buffers.

IvanMalison:
Fixes #52


IvanMalison:
@sshaw The other major issue is that vc won't necessary always use the full sha when creating the relevant buffer, so I think we will need to do some type of resolution on the ref that appears in between .~ .~

sshaw:
> The other major issue is that vc won't necessary always use the full sha 

That's fine. ~REV~ can be a tag, branch or SHA. As long as we take the ~REV~ part and pass it to the linking function we should be okay. All [supported services](https://github.com/sshaw/git-link#supported-services) can handle this.

sshaw:
Closing in lieu of 5d40a13.

Issue #49, opened 2018-02-02
Cannot parse tramp filenames


cswarth:
It would be great if this package could deal with tramp-mode filenames that look like `/ssh:aws-dev:/home/ubuntu/somedir/main.go`
When I called `git-link` I get the message `Can’t figure out what to link to`

sshaw:
Yes, very good idea. 

The problem seems to be with [running `git rev-parse --show-top-level`](https://github.com/sshaw/git-link/blob/472af75eca7aba42a1934f408c4bda55e34c9dcf/git-link.el#L198-L199). Your buffer is remote, but this is ran locally. 

Any idea how to fix?

sshaw:
It looks like [using `start-file-process`](https://www.gnu.org/software/emacs/manual/html_node/tramp/Remote-processes.html) will run on the remote server when buffer is a tramp file.

juergenhoetzel:
Had the same issue. Proposed fix: #54

PR #54, opened 2018-05-26
Tramp support

juergenhoetzel:
As already stated in #49, `process-lines` is not tramp-aware.

Also a *Tramp* `(buffer-file-name)` has to be dissected.

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -171,7 +171,15 @@ As an example, \"gitlab\" will match with both \"gitlab.com\" and
   :group 'git-link)
 
 (defun git-link--exec(&rest args)
-  (ignore-errors (apply 'process-lines `("git" ,@(when args args)))))
+  (ignore-errors
+    (with-temp-buffer
+      (when (zerop (apply #'process-file "git" nil (current-buffer) nil args))
+        (goto-char (point-min))
+        (cl-loop until (eobp)
+                 collect (buffer-substring-no-properties
+                          (line-beginning-position)
+                          (line-end-position))
+                 do (forward-line 1))))))
 
 (defun git-link--get-config (name)
   (car (git-link--exec "config" "--get" name)))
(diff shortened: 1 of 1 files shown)


sshaw:
Oh, very nice. Thank you. Give me a day or two to look into and play around with this. 👍

git-link.el
```diff
@@ -239,7 +247,8 @@ Return nil,
                       alist))))
 
 (defun git-link--relative-filename ()
-  (let* ((filename (buffer-file-name))
+  (let* ((filename (or (file-remote-p (buffer-file-name) 'localname)
```
sshaw: We support getting links from dired buffers and in this case `buffer-file-name` will be `nil` which will signal an error when passed to `file-or-remote-p`.

sshaw:
Also looks like the link is wrong if the buffer being visited via tramp is opened via a symlinked path.

juergenhoetzel:
> Also looks like the link is wrong if the buffer being visited via tramp is opened via a symlinked path.

Good catch: Also handled in the latest commit.

sshaw:
Hi, thanks. 

Calling `git-link` on tramp dired buffer produces wrong link. Maybe just need to call `(file-remote-p file 'localname) [here](https://github.com/sshaw/git-link/blob/a5a8a9e1c9ad1526b36fb3d3490b002bfdde6ec8/git-link.el#L251-L252)?


Also can you squash this into a single or logical commits?

juergenhoetzel:
> Hi, thanks.
> 
> Calling git-link on tramp dired buffer produces wrong link. Maybe just need to call (file-remote-p file 'localname) here?

I came up with a rather simple solution. There is no need to change `git-link--relative-filename`: It's already _Tramp-agnostic_.

I just adjusted `git-link--repo-root` to return tramp file names on remote hosts. In the latest (squashed) commit.

git-link.el
```diff
@@ -191,7 +199,10 @@ As an example, \"gitlab\" will match with both \"gitlab.com\" and
   (car (git-link--exec "symbolic-ref" "--short" "HEAD")))
 
 (defun git-link--repo-root ()
-  (car (git-link--exec "rev-parse" "--show-toplevel")))
+  (let ((dir (car (git-link--exec "rev-parse" "--show-toplevel"))))
+    (if (file-remote-p default-directory)
```
sshaw: Hi, the problem with this is that one can run `M-x cd` and change the default directory. Buffer remains as is, but repo root returned by this will be wrong.

juergenhoetzel: > Hi, the problem with this is that one can run M-x cd and change the default directory. Buffer remains as is, but repo root returned by this will be wrong.

Yes, but this case also fails on non-tramp buffers, because `git-link--exec` executes in `default-directory` so `git-link--repo-root` fails to get the correct root.

IMO this is a edge case no user will unintentional encounter.

sshaw: >  git-link--exec executes in default-directory

Hurmmm, yes. Ok, will try to somehow address this later then. 

Thanks again for this!

PR #58, opened 2018-10-25
Strip suffix on remote path only if it ends in .git

weshmashian:
Updated tests to reflect this.


weshmashian:
> Did you discover this while trying to link to your emacs.d repo? That's funny.

Yeah, pretty much. Thanks for merging this.

PR #60, opened 2019-07-05
git-link should work with magit-blame buffers

dcluna:
Hey, first of all, thanks for your work on this package, it's incredibly helpful.

So, I wrote some code to make this package return a github blame URL in case the user is in magit-blame-mode. I haven't really tried it with other hosts/handlers, but I believe it should work as-is, provided they work with commits and line numbers.

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -211,9 +211,10 @@ As an example, \"gitlab\" will match with both \"gitlab.com\" and
   (car (git-link--exec "--no-pager" "log" "-n1" "--pretty=format:%H")))
 
 (defun git-link--commit ()
-  (if (git-link--using-git-timemachine)
-      (car git-timemachine-revision)
-    (git-link--last-commit)))
+  (cond
+   ((git-link--using-git-timemachine) (car git-timemachine-revision))
+   ((git-link--using-magit-blame) (oref (magit-blame-chunk-at (point)) :orig-rev))
+   (t (git-link--last-commit))))
 
 (defun git-link--current-branch ()
   (car (git-link--exec "symbolic-ref" "--short" "HEAD")))
@@ -285,9 +286,14 @@ return (FILENAME . REVISION) otherwise nil."
       (cond
        ((eq major-mode 'dired-mode)
         (setq filename (dired-file-name-at-point)))
-       ((and (string-match-p "^magit-" (symbol-name major-mode))
-             (fboundp 'magit-file-at-point))
-        (setq filename (magit-file-at-point)))))
+       ((or
+         (git-link--using-magit-blame)
+         (string-match-p "^magit-" (symbol-name major-mode)))
+        (setq filename (or
+                        (and (fboundp 'magit-file-at-point)
+                             (magit-file-at-point))
+                        (and (fboundp 'magit-current-file)
+                             (magit-current-file)))))))
 
     (if (and dir filename
              ;; Make sure filename is not above dir, e.g. "/foo/repo-root/.."
@@ -328,6 +334,9 @@ return (FILENAME . REVISION) otherwise nil."
   (and (boundp 'git-timemachine-revision)
        git-timemachine-revision))
 
+(defun git-link--using-magit-blame ()
+  magit-blame-mode)
+
 (defun git-link--read-remote ()
   (let ((remotes (git-link--remotes))
 	(current (git-link--remote)))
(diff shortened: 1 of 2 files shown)


sshaw:
Hi, I'm glad you find the package useful. 

Thanks for the PR. I've been wanting this feature for a while! Will review.

git-link.el
```diff
@@ -328,6 +334,9 @@ return (FILENAME . REVISION) otherwise nil."
   (and (boundp 'git-timemachine-revision)
        git-timemachine-revision))
 
+(defun git-link--using-magit-blame ()
+  magit-blame-mode)
```
sshaw: This will result in an error if `magit-blame-mode` is not defined. You'll have to do something like [` git-link--using-git-timemachine`](https://github.com/sshaw/git-link/blob/1dbabfed4c5c3c5ac6ffa9035a9c3d4c6cc7f885/git-link.el#L327-L329).

git-link.el
```diff
@@ -211,9 +211,10 @@ As an example, \"gitlab\" will match with both \"gitlab.com\" and
   (car (git-link--exec "--no-pager" "log" "-n1" "--pretty=format:%H")))
 
 (defun git-link--commit ()
-  (if (git-link--using-git-timemachine)
-      (car git-timemachine-revision)
-    (git-link--last-commit)))
+  (cond
+   ((git-link--using-git-timemachine) (car git-timemachine-revision))
+   ((git-link--using-magit-blame) (oref (magit-blame-chunk-at (point)) :orig-rev))
```
sshaw: I don't see any documentation for `magit-blame-chunk-at`. Maybe it's not part of public interface? I see docs for `magit-blame-chunk-get`.  I'm using version `20171022.9`, maybe this is why?

git-link.el
```diff
@@ -285,9 +286,14 @@ return (FILENAME . REVISION) otherwise nil."
...
        ((eq major-mode 'dired-mode)
         (setq filename (dired-file-name-at-point)))
-       ((and (string-match-p "^magit-" (symbol-name major-mode))
-             (fboundp 'magit-file-at-point))
-        (setq filename (magit-file-at-point)))))
+       ((or
+         (git-link--using-magit-blame)
+         (string-match-p "^magit-" (symbol-name major-mode)))
+        (setq filename (or
+                        (and (fboundp 'magit-file-at-point)
```
sshaw: If we're in git blame mode`(fboundp 'magit-file-at-point)` will always be `t` but it will always return `nil`, right?  So it should only be executed when `(string-match-p "^magit-" (symbol-name major-mode))` is `t`. No need to always evaluate it

git-link.el
```diff
@@ -285,9 +286,14 @@ return (FILENAME . REVISION) otherwise nil."
...
+                             (magit-file-at-point))
+                        (and (fboundp 'magit-current-file)
+                             (magit-current-file)))))))
```
sshaw: And in a `^magit-` major mode this will always be bound  but will return `nil` unless we're in blame mode.

sshaw: Maybe the: 


```el
(or (git-link--using-magit-blame)
        (string-match-p "^magit-" (symbol-name major-mode)))
  (or (and (fboundp 'magit-file-at-point)
   ....
```
code makes sense if it's being used as a catchall for other  `^magit-` modes so that we make a best effort attempt to get a filename.

git-link.el
```diff
@@ -342,29 +351,31 @@ return (FILENAME . REVISION) otherwise nil."
...
-          (when (= end (line-beginning-position))
-            ;; Go up and avoid the blank line
-            (setq end (1- end)))
-          (setq line-end (line-number-at-pos end))
-          (when (<= line-end line-start)
-            (setq line-end nil)))
-        (list line-start line-end)))))
+  (cond
+   ((git-link--using-magit-blame) (list (oref (magit-blame-chunk-at (point)) :orig-line) nil))
+   (buffer-file-name (save-restriction
```
sshaw: Why are we calling `buffer-file-name` here? This is supposed to return start and end of region.

git-link.el
```diff
@@ -342,29 +351,31 @@ return (FILENAME . REVISION) otherwise nil."
...
-            (exchange-point-and-mark))
-          (when (= end (line-beginning-position))
-            ;; Go up and avoid the blank line
-            (setq end (1- end)))
-          (setq line-end (line-number-at-pos end))
-          (when (<= line-end line-start)
-            (setq line-end nil)))
-        (list line-start line-end)))))
+  (cond
+   ((git-link--using-magit-blame) (list (oref (magit-blame-chunk-at (point)) :orig-line) nil))
```
sshaw: Why not return the end of the selected region here? Of course we'd have to remove the lines added by magit-blame 😬

git-link.el
sshaw: Oh [I see why](https://github.com/sshaw/git-link/pull/60/files#diff-dbf604e34c6900c6747ecbe57dff621aL459). But the check of `buffer-file-name` variable must be done _before_ we attempt to get the region else there will be an error. 

When using magit-blame `buffer-file-name` is defined, so not sure why you removed it (Maybe it's not defined in your version? I know the latest magit made changes to its blame mode).

sshaw:
Closing due to lack of feedback and breaking functionality. Please feel free to reopen. This would be very nice to have.

git-link.el
dcluna: > I'm using version 20171022.9, maybe this is why?

Probably. I'm running magit from source, but `(magit-version)` returns `v2.90.1-701-gc9a754fa`. 

The corresponding release is this one: https://github.com/magit/magit/releases/tag/v2.90.1

It will be a bit difficult to support older magit versions - so this feature will probably be gated on `magit-blame-chunk-at` being available. Do you think this is a blocker for this PR?

> Maybe it's not part of public interface?

Yes, I didn't check that. When I update this PR (shouldn't take long) I'll take this into account. Thanks for the reminder!

PR #61, opened 2019-08-16
Add support for Magit-Blob buffers

Miciah:
Use the correct branch or commit when the current buffer is visiting a blob using Magit.

* `git-link.el` (`git-link--commit`): Use the `magit-buffer-revision` variable if the buffer is using Magit-Blob mode.
(`git-link--branch`): Look up the branch using `magit-buffer-revision` and the `magit-rev-branch` function if the buffer is using Magit-Blob mode.
(`git-link--relative-filename`): Use the `magit-buffer-file-name` variable the buffer is using Magit-Blob mode.
(`git-link--using-magit-blob-mode`): New function.
(`git-link`): Set the region and ignore the branch that is checked out into the working tree when using Magit-Blob mode.

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -211,9 +211,12 @@ As an example, \"gitlab\" will match with both \"gitlab.com\" and
   (car (git-link--exec "--no-pager" "log" "-n1" "--pretty=format:%H")))
 
 (defun git-link--commit ()
-  (if (git-link--using-git-timemachine)
-      (car git-timemachine-revision)
-    (git-link--last-commit)))
+  (cond
+   ((git-link--using-git-timemachine)
+    (car git-timemachine-revision))
+   ((git-link--using-magit-blob-mode)
+    magit-buffer-revision)
+   (t (git-link--last-commit))))
 
 (defun git-link--current-branch ()
   (car (git-link--exec "symbolic-ref" "--short" "HEAD")))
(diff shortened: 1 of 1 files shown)


git-link.el
```diff
@@ -233,6 +236,8 @@ As an example, \"gitlab\" will match with both \"gitlab.com\" and
 (defun git-link--branch ()
   (or (git-link--get-config "git-link.branch")
       git-link-default-branch
+      (when (git-link--using-magit-blob-mode)
+        (magit-rev-branch magit-buffer-revision))
```
sshaw: Under what magi command does this return a valid branch? I tried with a file opened from a commit via `magit-log` but no branch.

Miciah: I open a blob from a branch using `magit-find-file`, and then `(magit-rev-branch magit-buffer-revision)` evaluates to the name of that branch.

You likely opened a blob that is not from the tip of a branch, in which case `magit-buffer-revision` will have the right commit id, but `(magit-rev-branch magit-buffer-revision)` will evaluate to nil.  In this case, I believe the right thing to do would be to return the commit id, but I see that `git-link` is in fact using the checked-out branch.  I will amend my changes to fix this problem.

Miciah:
> I noticed that it does not include the line numbers when linking from a blob.

Thanks for noticing this, I had missed it.  I'll look into it.

Miciah:
I believe I fixed the missing line numbers.  (I'm curious why the `when` form that I modified is needed at all.)

sshaw:
Great, thanks again!

sshaw:
> (I'm curious why the when form that I modified is needed at all.)

Because one can link to a file in a dired buffer or in magit buffers and in these case we don't have an actual file and therefore cannot/should not get the region.

PR #62, opened 2019-08-21
Add support for Azure repositories

r-darwish:


sshaw:
Hi, thanks for this nice addition. I'm excited to get it added.

When the repository is cloned via SSH the wrong URL is used. For example:

```
/tmp/Test >git remote -v
origin	s-shaw@vs-ssh.visualstudio.com:v3/s-shaw/Test/Test (fetch)
origin	s-shaw@vs-ssh.visualstudio.com:v3/s-shaw/Test/Test (push)
```

Results in:

https://vs-ssh.visualstudio.com/v3/s-shaw/Test/Test?path=foo.txt&version=GBmaster&line=2

Which results in a 404.

git-link.el
```diff
@@ -166,7 +166,8 @@
     ("github" git-link-github)
     ("bitbucket" git-link-bitbucket)
     ("gitorious" git-link-gitorious)
-    ("gitlab" git-link-gitlab))
+    ("gitlab" git-link-gitlab)
+    ("visualstudio\\|azure" git-link-azure))
```
sshaw: Under what circumstances would one have an azure domain name?

git-link.el
```diff
@@ -411,12 +413,30 @@ return (FILENAME . REVISION) otherwise nil."
...
 	  hostname
 	  dirname
 	  commit))
 
+(defun git-link-commit-azure (hostname dirname commit)
+  (format "https://%s/%s/commit/%s"
+	  hostname
+	  dirname
+	  (with-temp-buffer
+        (vc-git--out-ok "rev-parse" commit)
```
sshaw: What's the reason for calling `vc-git--out-ok` here? This is not done elsewhere in the existing code and it (seems) to work fine without it when linking to commits.

sshaw: Also given emacs conventions `vc-git--out-ok` is not part of the public API.

r-darwish:
> Hi, thanks for this nice addition. I'm excited to get it added.
> 
> When the repository is cloned via SSH the wrong URL is used. For example:
> 
> ```
> /tmp/Test >git remote -v
> origin	s-shaw@vs-ssh.visualstudio.com:v3/s-shaw/Test/Test (fetch)
> origin	s-shaw@vs-ssh.visualstudio.com:v3/s-shaw/Test/Test (push)
> ```
> 
> Results in:
> 
> [vs-ssh.visualstudio.com/v3/s-shaw/Test/Test?path=foo.txt&version=GBmaster&line=2](https://vs-ssh.visualstudio.com/v3/s-shaw/Test/Test?path=foo.txt&version=GBmaster&line=2)
> 
> Which results in a 404.

Good catch. I only checked HTTP. I'll probably need to add some kind on manipulation when the link is SSH. Could you tell me that is the HTTP URL of your repository? I think that the SSH user should replace `vs-ssh`

git-link.el
r-darwish: I use it to get the full commit hash as the Azure interface seem not to handle short commits well.

I usually run git-link from the magit log buffer. The hash that appear in that buffer is short.

Do you have another suggestion for getting the full commit hash?

git-link.el
r-darwish: When creating a repository from Azure pipelines, such as this:

https://dev.azure.com/r-darwish/_git/testing

r-darwish:
Thanks for replying quickly. I'll probably be able to commit the fixes in Sunday.

sshaw:
> Could you tell me that is the HTTP URL of your repository? I think that the SSH user should replace vs-ssh

https://s-shaw.visualstudio.com/Test

git-link.el
sshaw: Oh I see, these are more or less interchangeable:

https://dev.azure.com/s-shaw/_git/Test == https://s-shaw.visualstudio.com/_git/Test

git-link.el
sshaw: > Do you have another suggestion for getting the full commit hash?

I will investigate.

r-darwish:
@sshaw I pushed a commit which normalizes the SSH links. Sorry if my elisp is horrible. This is my first non-trivial contribution for an elisp package.

git-link.el
sshaw: Thinking back and looking though Emacs' docs –and this code– I recall that at one point I thought: "I should have just used Emacs's `vc-xxx` [or whatever] functions to interact with git" but then I found that there isn't a well-defined public API for this type of usage.   

The way I have done things like this is to make a function to call `git`. There are [several examples of this](https://github.com/sshaw/git-link/blob/5581b4760aa2efb01eddca035a2758bc62da8635/git-link.el#L209) in the code now.

git-link.el
```diff
@@ -411,12 +428,30 @@ return (FILENAME . REVISION) otherwise nil."
...
+        (string-trim (buffer-string)))))
```
sshaw: Why is the call to `buffer-string` needed? The functions should depend on their args and not on the caller's context.

r-darwish:
Done 🙂

sshaw:
Great, thanks so much!

sshaw:
Released in [v0.7.5](https://github.com/sshaw/git-link/releases/tag/v0.7.5)

PR #63, opened 2019-09-25
Define function to get absolute filename

lafrenierejm:
This provides a cleaner, more functional approach than the existing `setq`-heavy
implementation.  Additionally, a fallback of `default-directory` (which is
guaranteed to exist) will be used if no other method of obtaining the filename
is valid.


sshaw:
Hi, thanks. 

Before `filename` defaulted to `(buffer-file-name)` and only used the `cond` to get the file's name if  `(buffer-file-name)` was `nil`. Now `(buffer-file-name)` is never used and one cannot get a to file file in the buffer

lafrenierejm:
Thank you for catching that oversight.  It should be fixed now.

lafrenierejm:
@sshaw Would you like me to squash my commits?

sshaw:
Hi yes please squash but... let me test first. Give me a day. Sadly I have no automated tests here. 😢

git-link.el
```diff
@@ -290,20 +290,30 @@ return (FILENAME . REVISION) otherwise nil."
...
+   ;; Else if Magit knows the file...
+   ((and (git-link--using-magit-blob-mode)
+         magit-buffer-file-name))
+   ;; Else if over a file in a Magit buffer...
+   ((and (string-match-p "^magit-" (symbol-name major-mode))
+         (fboundp 'magit-file-at-point)
+         (magit-file-at-point)))
+   ;; Else...
+   (t
+    default-directory)))
```
sshaw: This causes buffers without files to now link to the repo root instead of displaying a "can't figure out what to link to error"

This function also changes the old semantics in that if say `(dired-file-name-at-point)` returns `nil` the next case is processed. Whereas before if it returned `nil` with stopped the mode based processing. 

Really I don't think this I see the need for this function.

lafrenierejm: > This function also changes the old semantics in that if say (dired-file-name-at-point) returns nil the next case is processed. Whereas before if it returned nil with stopped the mode based processing.

That is intentional.  I was finding myself running `git-link` while on the top line of a Dired buffer expecting a URL of the directory but instead receiving an error.

lafrenierejm: > This causes buffers without files to now link to the repo root instead of displaying a "can't figure out what to link to error"

Indeed that is undesirable.  I have found it to be worth it to be able to link to
- the repo root from a generic position in a Magit status buffer and
- the directory itself from a Dired buffer (as discussed above).

It definitely falls short for the use-case you're describing, where the user has started a file they intend to commit but have yet to commit/push.

lafrenierejm: > Really I don't think this I see the need for this function.

That's totally fair.  I would be glad to
- work on addressing the case of non-Dired, non-Magit buffers to prevent unintended use-cases;
- open a new PR that is just the code cleanup without any changes in functionality; or
- drop this PR entirely.

No hurt feelings whatever you decide. :)

sshaw: > I was finding myself running git-link while on the top line of a Dired buffer expecting a URL of the directory but instead receiving an error.

Hurmm, I see. Let me check this out. 

>  I have found it to be worth it to be able to link to
> the repo root from a generic position in a Magit status buffer and
> the directory itself from a Dired buffer (as discussed above).

Makes sense. There is also a `git-link-homepage` function. But this does not include the branch in the URL.

Let me look again given this new info. Thanks!

lafrenierejm: @sshaw What would you think of falling back to `default-directory` only for Dired and Magit buffers?

sshaw:
@lafrenierejm I have not forgotten. Will review in Nov!

git-link.el
sshaw: Hi, 

With Magit, the buffers can be more complex. One may _think_ they're on file X, but they're not. With the recommended addition we would just give them a link, even if it's not the one they wanted, so falling back to `default-directory` may be confusing. 

With Dired  it _seems_ to makes sense to fall back to `default-directory` because here we're explicitly browsing a directory and the buffer is all files so,  if nothing is at point, it seems like a reasonable default **but** this can suffer from the Magit confusion scenario if one uses `dired-maybe-insert-subdir`. In this case, the Dired buffer can have more than 1 directory, so `default-directory` would be "wrong" if point was on/under the inserted subdir. If we could somehow determine what "directory scope" point is at, that would it make sense.

lafrenierejm: > With Magit, the buffers can be more complex. One may think they're on file X, but they're not. With the recommended addition we would just give them a link, even if it's not the one they wanted, so falling back to default-directory may be confusing.

Can you elaborate on this?  I have only experienced cases where Magit buffers either have no `buffer-file-name` or `buffer-file-name` points to the correct file.

lafrenierejm: > With Dired it seems to makes sense to fall back to default-directory because here we're explicitly browsing a directory and the buffer is all files so, if nothing is at point, it seems like a reasonable default but this can suffer from the Magit confusion scenario if one uses dired-maybe-insert-subdir. In this case, the Dired buffer can have more than 1 directory, so default-directory would be "wrong" if point was on/under the inserted subdir. If we could somehow determine what "directory scope" point is at, that would it make sense.

I had never heard of `dired-maybe-insert-subdir` before.  I will look into it and reply back with my findings for how it interacts with `default-directory`.

sshaw: >> With Magit, the buffers can be more complex. One may think they're on file X, but they're not. 

> Can you elaborate on this?

`magit-status` and `magit-show-commit` come to mind. These are commonly used and usually opened via typical Magit navigation. Defaulting to `default-directory`'s contents seems odd if one clicked anywhere here.

lafrenierejm:
@sshaw What would you think of my just removing the `default-directory` fallback for now? I can revisit that in a separate PR.

sshaw:
Hi, sorry missed your last comment. Give me 24 hours to get caught up. Feel free to harass me if I don't get back to you... ⏳

sshaw:
> What would you think of my just removing the default-directory fallback for now? 

Sounds good 👍. Please squash into a single commit, thanks.

sshaw:
Closing. Feel free to reopen with updated changes.

PR #65, opened 2019-10-31
Adapt to changes in Azure interface

r-darwish:
The new interface of Azure requires the following changes:

* File names must start with `/`
* `line` and `lineEnd` must be specified
* `lineStartColumn` and `lineEndColumn` must be specified. Since git-link does not support linking by
  column then these values are always set to 1 and 9999

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -432,14 +432,13 @@ return (FILENAME . REVISION) otherwise nil."
                               (format "L%s" start)))))))
 
 (defun git-link-azure (hostname dirname filename branch commit start end)
-  (format "https://%s/%s?path=%s&version=%s%s"
+  (format "https://%s/%s?path=%%2F%s&version=%s&line=%s&lineEnd=%s&lineStartColumn=1&lineEndColumn=9999&lineStyle=plain"
 	  hostname
 	  dirname
       filename
       (concat "G" (if branch "B" "C") (or branch commit))
-      (concat
-       (when start (format "&line=%s" start))
-       (when end (format "&lineEnd=%s" end)))))
+      start
+      (or end start)))
 
 (defun git-link-commit-github (hostname dirname commit)
   (format "https://%s/%s/commit/%s"


sshaw:
Needs to be changed already ‽

> File names must start with /

Do you mean end?

> line and lineEnd must be specified

So when one is linking to a single line the params value is empty? In the current PR does it link as empty string or `"nil"`?

> lineStartColumn and lineEndColumn 

Hurmmm. Intriguing. Going to have to check these links out.

Is there a reason to not provide columns? May be a good feature (if not now them at some point).

r-darwish:
> Needs to be changed already ‽

Microsoft just deployed a new UI. I don't think it means that it has to be changed on a regular basis.

> Do you mean end?

No, I actually mean start. If a file in the repository called `dirname/filename` then it has to be specified as `/dirname/filename`.

> So when one is linking to a single line the param's value is empty? 

When using the Azure interface itself to select line N, the interface will add `lineEnd=N+1`. This is also what the PR does.

> Is there a reason to not provide columns?

That could be nice, but since only Azure supports it I figured it would be too esoteric to bother, as it requires changing the signature of all functions (or use optional parameters)

git-link.el
```diff
@@ -432,14 +432,15 @@ return (FILENAME . REVISION) otherwise nil."
                               (format "L%s" start)))))))
 
 (defun git-link-azure (hostname dirname filename branch commit start end)
-  (format "https://%s/%s?path=%s&version=%s%s"
+  (format "https://%s/%s?path=%s&version=%s&line=%s&lineEnd=%s&lineStartColumn=1&lineEndColumn=9999&lineStyle=plain"
 	  hostname
 	  dirname
-      filename
+      (if (not (string-prefix-p "/" filename))
+          (concat "/" filename)
```
sshaw: I don't think `filename` will ever begin with a forward slash. Looking at existing handlers, they all assume it wont be there. Have you seen otherwise? If not, I think we can always include it. But note that this should be URL encoded (this may be an existing bug with other handlers) so maybe just add `path=%27%s` to the `format` template?

sshaw:
>> Needs to be changed already ‽

> Microsoft just deployed a new UI. I don't think it means that it has to be changed on a regular basis. The new parameters were there before, but they just became mandatory

The old URLs still work. 

>> Is there a reason to not provide columns?

> That could be nice, but since only Azure supports it I figured it would be too esoteric to bother, as it requires changing the signature of all functions (or use optional parameters)

I think this is key feature, linking to columns. But links do work without them (using a test repo on dev.azure.com). Looks like maybe modifying `git-link--get-region` to return this info would be the way to go, then just adding –as you've said– an optional alist argument with provider specific info. 

Either way,  checkout my comment about `filename`. Afterwards should be good to merge.

sshaw:
Great thanks! Please squash into a single commit.

r-darwish:
Done. Note that Github itself [supports squashing a pull request](https://github.blog/2016-04-01-squash-your-commits/)

sshaw:
Squashing PRs via GitHub is a bad thing as the original SHA is lost.

PR #67, opened 2020-03-21
try use tag/commit as branch when detached

jiacai2050:
get current tag with reference to https://stackoverflow.com/a/18660163/2163429

I use golang/tools to test:
- https://github.com/golang/tools/blob/gopls%2Fv0.3.4/internal/lsp/mod/hover.go#L42

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -164,6 +164,15 @@
   :type 'boolean
   :group 'git-link)
 
+(defcustom git-link-preferred-format '(branch
+                                       tag
+                                       commit)
+  "Priority of format style used in URL"
+  :type '(repeat (choice (const branch)
+                         (const tag)
+                         (const commit)))
+  :group 'git-link)
+
 (defcustom git-link-remote-alist
   '(("git.sr.ht" git-link-sourcehut)
     ("github" git-link-github)
@@ -229,6 +238,9 @@ As an example, \"gitlab\" will match with both \"gitlab.com\" and
 (defun git-link--current-branch ()
   (car (git-link--exec "symbolic-ref" "--short" "HEAD")))
 
+(defun git-link--current-tag ()
+  (car (git-link--exec "describe" "--tags" "--exact-match")))
+
 (defun git-link--repo-root ()
   (let ((dir (car (git-link--exec "rev-parse" "--show-toplevel"))))
     (if (file-remote-p default-directory)
@@ -247,6 +259,11 @@ As an example, \"gitlab\" will match with both \"gitlab.com\" and
       (when (git-link--using-magit-blob-mode)
         (magit-rev-branch magit-buffer-revision))
       (git-link--current-branch)))
+
+(defun git-link--tag ()
+  (or (when (git-link--using-magit-blob-mode)
+        (magit-tag-at-point))
+      (git-link--current-tag)))
 
 (defun git-link--remote ()
   (let* ((branch (git-link--current-branch))
@@ -436,7 +453,7 @@ return (FILENAME . REVISION) otherwise nil."
 	  hostname
 	  dirname
       filename
-      (concat "G" (if branch "B" "C") (or branch commit))
+      branch
       start
       (or end start)))
 
@@ -496,6 +513,12 @@ return (FILENAME . REVISION) otherwise nil."
       (git-link--read-remote)
     (git-link--remote)))
 
+(defun git-link--prepare-prefix (host format placeholder)
+  (if (or (string= "dev.azure.com" host)
+          (string-suffix-p ".visualstudio.com" host))
+      (concat "G" (if (eq 'branch format) "B" "C") placeholder)
+     placeholder))
+
 ;;;###autoload
 (defun git-link (remote start end)
   "Create a URL representing the current buffer's location in its
@@ -515,7 +538,6 @@ Defaults to \"origin\"."
 
       (setq remote-info (git-link--parse-remote remote-url)
             filename    (git-link--relative-filename)
-            branch      (git-link--branch)
             commit      (git-link--commit)
             handler     (git-link--handler git-link-remote-alist (car remote-info)))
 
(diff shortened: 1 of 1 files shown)


sshaw:
Hi, thanks for this. Good idea. 

So without your commit and without setting `git-link-use-commit` to `t`, you were getting an empty path segment, correct e.g., `some//path/foo`? 

I think this could be turned into a feature/variable that instructs git-link to _always_ use a tag? Any desire to try this? If not no problem, just please confirm the above path issue for me.

Thanks a lot!

jiacai2050:
> So without your commit and without setting git-link-use-commit to t, you were getting an empty path segment, correct e.g., some//path/foo?

Yes, it looks like this
`https://github.com/golang/tools/blob//internal/lsp/mod/hover.go#L42`

But you remind me to deal with detached  state without any tag on it, if then, just use commit.

sshaw:
> So without your commit and without setting git-link-use-commit to t, you were getting an empty path segment, correct e.g., some//path/foo?

I've created #68 to address this. 

> But you remind me to deal with detached state without any tag on it, if then, just use commit.

This is not needed. We only want to use the commit _if_ `git-link-use-commit` is set to `t`.

This is how I think the `git-link` function should behave: 

1. Fix #68 
1. Add a variable `git-link-use-tag` defaulting to `nil`
1. Look up branch as we are now
1. Look up commit as we are now
1. Look up tag the tag [here](https://github.com/sshaw/git-link/blob/2fb0e4be8801e7d53e02fc7f14ec0f9a14586ba2/git-link.el#L518-L519)
1. Given a detached head state:
    1. **If** `git-link-use-commit` is `nil` and  `git-link-use-tag` is `nil` signal an error
    1. **Else** use whatever is appropriate based on the above variables

But with the variable `git-link-use-tag`, this should _always link with a tag when set_. Agree?

In a future version `git-link-use-commit` will default to `t`. Then you will not have these problems by default. It was a mistake to not make this the default from v0.0.1.

sshaw:
Assuming you agree, would you be okay with modifying the PR to work with steps 2, 5, and 6?

jiacai2050:
I don't think we should raise an error in step 6.1. 
From users' points of view, this package should work out of box wihout any configuration, they can set `git-link-use-tag` to `t` if they want permalink, but when HEAD is detached, this package should try to guess what is most suitable here, and use tag or commit id as fallback(with tag preferred).

> But with the variable git-link-use-tag, this should always link with a tag when set. Agree?

I agree with this. However if no tag is attached on HEAD, git-link should use branch/commit as fallback (with branch prefered).

If both `git-link-use-tag` and `git-link-use-commit` are set to `t`, then try tag first, fallback to commit when failed.

Any thoughts?

sshaw:
Hi, thanks for the feedback.

> I don't think we should raise an error in step 6.1.
> From users' points of view, this package should work out of box without any configuration,

I mostly agree, which is why at some point, i.e., v1, `git-link-use-commit` should default to `t`.
But from here on let's assume we're talking about a v1.

> but when HEAD is detached, this package should try to guess what is most suitable here, and use tag or commit id as fallback (with tag preferred).

If someone has `git-link-use-commit` set to `nil` it means they don't want a commit SHA to be used in a link. If a link is then generated with a commit, to me, this is a bug. Same applies to `git-link-use-tag` being `nil` and getting a link with a `tag`. 

> But with the variable git-link-use-tag, this should always link with a tag when set. Agree?

>> I agree with this. However if no tag is attached on HEAD, git-link should use branch/commit as fallback (with branch preferred).

Branch preferred by who? This is where the options come in. People have different preferences and one may not like the generated link if we don't respect their preferences/settings.

> If both git-link-use-tag and git-link-use-commit are set to t, then try tag first, fallback to commit when failed.
> Any thoughts?

While it does make things more complex implementation-wise, maybe the solution is to add, in addition to `git-link-use-tag`:

1. `git-link-preferred-format` which could be a list from most to least preferred: `(tag commit branch)` or similar
1. `git-link-no-matter-what` which if `t` will disregard all settings and attempt to generate a link based on `git-link-fallback-order`.

Though `git-link-preferred-format` maybe could replace `git-link-use-commit`, etc... Though I'd probably keep and deprecate them until a v2 (if there's ever a v2).

jiacai2050:
I like `git-link-preferred-format` idea, the value can be a list of function, then users can return what they want in the url.

I push another commit, please take another look.

git-link.el
```diff
@@ -164,6 +164,15 @@
   :type 'boolean
   :group 'git-link)
 
+(defcustom git-link-preferred-format '((lambda ()
+                                         (url-hexify-string (git-link--branch)))
+                                       (lambda ()
+                                         (url-hexify-string (git-link--tag)))
+                                       git-link--commit)
```
sshaw: What's the benefit to a list of functions vs just a symbol, e.g., `branch`, `tag`..? `git-link--branch` etc... are internal functions so caller should not use them.

sshaw:
```el
(defcustom git-link-preferred-format '(branch tags commit))
; ...
(funcall (intern format "git-link--%s" style)))
```
```

File: git-link.el [hunk 2]
```diff
@@ -496,6 +513,11 @@ return (FILENAME . REVISION) otherwise nil."
       (git-link--read-remote)
     (git-link--remote)))
 
+(defun empty-string-p (string)
+  "Return true if the STRING is empty or nil. Expects string type."
+  (or (null string)
+      (zerop (length (string-trim string)))))
```
sshaw: Is there a need to trim? If not maybe use `string-empty-p`? I had tried to avoid for compatability but at this point maybe safe to require minimum of 24.4?

git-link.el
jiacai2050: DONE.

git-link.el
jiacai2050: DONE

sshaw:
Hi, thanks. Sorry for the delay. It's a bit of a pain to test this across all supported services. The tests cases are non existent here :(

Nevertheless a few things: 

* I pointed the PR at the version-1.0 branch. As discussed, this will be part of v1 release and I will have to:
  * Deprecate + add support for the existing config variables
  * Continuing to support local `git config` stuff   
* Please squash into a single commit
* Linking to an Azure repository does not work, it has to do with [this](https://github.com/sshaw/git-link/blob/1d47714cb80e2bb434bf1ca4ab078b04aa86e21c/git-link.el#L442): it's being set to `B` even when a commit is the preferred format. This results in a 404.

jiacai2050:
Hi, maintaining compatibility can be a pain thing, and I have taken carefully not to break any existing options, please take another look, I have fixed the Azure issue.

git-link.el
```diff
@@ -496,6 +513,12 @@ return (FILENAME . REVISION) otherwise nil."
       (git-link--read-remote)
     (git-link--remote)))
 
+(defun git-link--prepare-prefix (host format placeholder)
+  (if (or (string= "dev.azure.com" host)
+          (string-suffix-p ".visualstudio.com" host))
+      (concat "G" (if (eq 'branch format) "B" "C") placeholder)
+     placeholder))
```
sshaw, line 520: Hi, 

Endpoint specific handling should be encapsulated by the endpoint specific function. With the addition of `git-link--prepare-prefix`, the Azure specific handling is no longer encapsulated by `git-link-azure`. As a consequence, we have a new issue:

The above function does not consider `git-link-commit-remote-alist`. With this, one can direct `git-link-azure` be called  for any host, not just the defaults. So if one does:

```el
(add-to-list 'git-link-remote-alist '("somehost\\.com" git-link-azure)) 
```

The `git-link` function will still fail.

The new link generating conditionals already seems to have a lot of "gotchas": hence the checks using `memq`, `string-empty-p` and `null` (I do realize that this is in part do to excising in consistencies with return values). Now we're adding `throw` and `catch` and `dolist`. We've gone from 1 line to 7 or so.

I appreciate your work and I hate not being able to merge. The whole ideal of a preferred format list needs to be addressed holistically, from a higher level. It's clear that when we attempt to wedge it into the existing framework for link generation problems occur and code becomes more complex than it should be.

sshaw: My lack of testing here is also a problem. With proper tests, you'd get this feedback immediately instead of having to wait days for me :(

jiacai2050: It seems git-link need a refactor to solve your concerns, I'm unsure I'm on the right way, maybe your can draft a detailed design, or just implement it based on my changes.

sshaw: Yeah it's unfortunate but at least this list business came out of it which is a nice idea. Will do a refactor a some point for a 1.0 which is definitely overdue.

jiacai2050: Thanks for your hard work on this.

Issue #45, opened 2017-07-26
use user-error to signal errors


wbolster:
first, thanks for this package!

that said, please use `(user-error ...)` instead or `(message ...)` to signal errors, so that i can catch them and do something sensible.

this would allow me to write a function that would first try `(git-link-commit)`, falling back to `(git-link)` in case that fails, which would basically "do what i mean" in 99% of my use cases.

let me know whether i should submit a pr!

sshaw:
Hi, I only call `error` [in `git-link-homepage`](https://github.com/sshaw/git-link/blob/master/git-link.el#L470). `messages` is most everywhere. 

Can you provide more info? Thanks!

wbolster:
yeah, my suggestion is to use `(user-error)` instead of `(message)` when something goes wrong. right now you use `(message)` which does not trigger an error that can be caught.

i want to be able to do something like this:

``` elisp
    (condition-case nil
        (git-link-commit)
      (user-error
       (git-link)))
```

but it seems right now i cannot.

sshaw:
Got it, makes sense. I'll have to look into user-error. 

You may also want to checkout the PRs and issues. There's talk of making a `git-link-dwim` function or maybe even changing `git-link` to fallback to `git-link-commit`. 

I think changing these functions to signal errors will require releasing a 1.0 version.

wbolster:
hey @sshaw 👋🏼, long time no see 🙃

i recently noticed that `git-link` doesn't report errors correctly, looked at the issue tracker, and to my own surprise, it was _me_ who first opened this issue… back in 2017! 🤯

let's get it fixed once and for all. i just opened #154, so please take a look!

PR #69, opened 2020-04-16
Create a function to return the git-link itself

futuro:
`git-link` is incredibly useful, but it doesn't return the URL it
creates. This introduces a new function which does so, allowing more
tooling to make use of this great feature.

This builds from the conversation in #38 and issue #37.

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -354,8 +354,8 @@ return (FILENAME . REVISION) otherwise nil."
 
 (defvar magit-buffer-file-name)
 
-(defun git-link--relative-filename ()
-  (let* ((filename (buffer-file-name (buffer-base-buffer)))
+(defun git-link--relative-filename (&optional filename)
+  (let* ((filename (or filename (buffer-file-name (buffer-base-buffer))))
 	 (dir      (git-link--repo-root)))
 
     (when (null filename)
@@ -666,6 +666,55 @@ return (FILENAME . REVISION) otherwise nil."
     (git-link--remote)))
 
 ;;;###autoload
+(defun git-link-url (file-or-buffer remote &optional start end use-branch-p)
+  "Create a URL for the FILE-OR-BUFFER's location at REMOTE.
+
+When given START and END, include the current line number or
+active region.
+
+When given USE-BRANCH-P, force the use of the branch name in the
+resultant url."
+  (let* ((filename (git-link--relative-filename
+                    (if (bufferp file-or-buffer)
+                        (buffer-file-name file-or-buffer)
+                      file-or-buffer)))
+         (remote      (or remote git-link-default-remote))
+         (remote-url  (git-link--remote-url remote))
+         (remote-info (git-link--parse-remote remote-url))
+         (branch      (git-link--branch))
+         (commit      (git-link--commit))
+         (handler     (git-link--handler git-link-remote-alist (car remote-info))))
+    (cond ((null remote-url)
+           (user-error "Remote `%s' not found" remote))
+          ((null filename)
+           (user-error "Can't figure out what to link to"))
+          ((null (car remote-info))
+           (user-error "Remote `%s' contains an unsupported URL" remote))
+          ((not (functionp handler))
+           (user-error "No handler found for %s" (car remote-info)))
+          ;; TODO: null ret val
+          (t
+           (let* ((vc-revision (git-link--parse-vc-revision filename))
+                  (hostname (car remote-info))
+                  (dirname (cadr remote-info))
+                  (branch (if (and (not use-branch-p)
+                                   (or (git-link--using-git-timemachine)
+                                       (git-link--using-magit-blob-mode)
+                                       vc-revision
+                                       git-link-use-commit))
+                              nil
+                            (url-hexify-string branch)))
+                  (filename (if vc-revision (car vc-revision) filename))
+                  (commit (if vc-revision (car vc-revision) commit)))
+             (funcall handler
+                      hostname
+                      dirname
+                      filename
+                      branch
+                      commit
+                      start
+                      end))))))
+
 (defun git-link (remote start end)
   "Create a URL representing the current buffer's location in its
 GitHub/Bitbucket/GitLab/... repository at the current line number
@@ -688,43 +737,10 @@ Defaults to \"origin\"."
            (list remote nil nil)
          (list remote (car region) (cadr region))))))
 
-  (let (filename branch commit handler remote-info (remote-url (git-link--remote-url remote)))
-    (if (null remote-url)
-        (message "Remote `%s' not found" remote)
-
-      (setq remote-info (git-link--parse-remote remote-url)
-            filename    (git-link--relative-filename)
-            branch      (git-link--branch)
-            commit      (git-link--commit)
-            handler     (git-link--handler git-link-remote-alist (car remote-info)))
-
-      (cond ((null filename)
-             (message "Can't figure out what to link to"))
-            ((null (car remote-info))
-             (message "Remote `%s' contains an unsupported URL" remote))
-            ((not (functionp handler))
-             (message "No handler found for %s" (car remote-info)))
-            ;; TODO: null ret val
-            (t
-             (let ((vc-revison (git-link--parse-vc-revision filename)))
-               (when vc-revison
-                 (setq filename (car vc-revison)
-                       commit   (cdr vc-revison)))
-
-               (git-link--new
-                (funcall handler
-                         (car remote-info)
-                         (cadr remote-info)
-                         filename
-                         (if (or (git-link--using-git-timemachine)
-                                 (git-link--using-magit-blob-mode)
-                                 vc-revison
-                                 git-link-use-commit)
-                             nil
-                           (url-hexify-string branch))
-                         commit
-                         start
-                         end))))))))
+  (condition-case err
+      (git-link--new (git-link-url buffer-file-name remote start end))
+    (user-error
+     (message (cadr err)))))
 
 ;;;###autoload
 (defun git-link-commit (remote)


sshaw:
Hi @futuro, thanks for the PR.  

This is definitely something that's needed. I believe my concern with the other PRs was compatibility in that the signatures and/or return value must be changed in order to have a solid function _and_ that calling `git-link` or whatever non-interactively should not add to the kill ring. 

This certainly avoids that (yay) but what about the rest of the global state? For example, if you're using if for an org capture template then won't all the checks `git-link-url` is doing on the current buffer produce wrong or non-deterministic results?

futuro:
Hey @sshaw,

I'll admit that I'm totally ignorant about global state (and emacs hacking in general), so I don't really have a good answer for this. I'll say that the function which is calling `git-link-url` is wrapped inside a `with-current-buffer`, so it's running (from what I understand) inside of the buffer for the file I want the git link for.

I'd welcome any pointers on what to read to understand the global state that's at play, or idiomatic design patterns for issues like this.

futuro:
Thinking about this more, @sshaw is your question about whether `git-link-url` can be called as a pure function, versus _needing_ to be called inside the buffer you want the link from?

sshaw:
Hi @futuro, sorry for the holdup

>  is your question about whether git-link-url can be called as a pure function, versus needing to be called inside the buffer you want the link from?

Yes, I think that a standalone function should be used. Something like:

```el
(git-link-url file-or-buffer &optional start end remote use-branch)
```

dotemacs:
Hi @sshaw, 

Thanks for the super useful package!

Is there a blocker for this feature to be merged? Thanks

sshaw:
> Is there a blocker for this feature to be merged? Thanks


It does not implement [a standalone version](https://github.com/sshaw/git-link/pull/69#issuecomment-631768670) of `git-link-url` (or similar). In its current form its return value depends on global state (unless I'm missing something?).

futuro:
I'll see if I can shake that out today.

futuro:
While I sort this out, @sshaw, I've noticed that `git-link` works in the same way as my change -- my change is just to extract some functionality from `git-link` -- and that much of `git-link` depends on global state, such as with an interactive call or within a `with-current-buffer` body.

My current thought is to have `git-link-url` contain a `with-current-buffer` call and have everything else be within that macro, but I'd love your thoughts on this and whether it's violating some kind of elisp design principle.

sshaw:
Hi, 

Thanks for getting back to this!

>  I've noticed that git-link works in the same way as my change  ...
>  and that much of git-link depends on global state ...

The default case for git-link for interactive use, i.e., get a link of/based on the current buffer's contents and cursor position. This is okay. 

>  My current thought is to have git-link-url contain a with-current-buffer call 
> ... but I'd love your thoughts on this and whether it's violating some kind of elisp design principle.

This is where the problem lies. If it is going to be a legit stand-alone function it must not depend on things like the current buffer, cursor position, etc... Instead of being for interactive use it should have a signature similar to that outlined [here](https://github.com/sshaw/git-link/pull/69#issuecomment-631768670) and must return the URL and not add it to the kill ring.

I would say this is general principle for functions, etc...: don't depend on global state.

Once we have this baseline function it can be used for all use cases, including the interactive one. 

Does that help?

sshaw:
> I would say this is general principle for functions, etc...: don't depend on global state.

Though we can depend on the output of `git`, which is sorta global state. If we didn't then we'd have more function arguments and I think it would loose some of its utility forcing the caller to pass in parsed `git` output.

futuro:
@sshaw after reviewing more of the code, I think I'm starting to see the way this could work. One question: what's the goal of the optional arg `use-branch`?

sshaw:
> ... what's the goal of the optional arg use-branch?

That is supposed to be the equivalent of sorts of [`git-link-use-commit`](https://github.com/sshaw/git-link#git-link-use-commit). Currently the default is to generate the link using the branch but time has shown that it's better to use the commit. So I think eventually `git-link-use-commit` will be `git-link-use-branch` to better reflect this.

futuro:
> That is supposed to be the equivalent of sorts of git-link-use-commit.

Rockin, that makes sense.

Another question! You'd put the remote as an optional argument, but the function can't work without a specified remote (as far as I understand it, at least). I'm thinking about moving this to be a required argument; am I missing anything?

futuro:
Ok @sshaw, this should be ready.

I went ahead with making `remote` a required argument, but let me know if I missed something and it should be optional.

futuro:
J/k, sorry. I need to fix a void-variable reference. Will ping you once it's solid.

futuro:
Ok, tested, and should be good to go for review @sshaw.

sshaw:
Thanks a lot. I will have a looksy at or before Christmas Day 🎅🤶

git-link.el
```diff
@@ -666,6 +666,54 @@ return (FILENAME . REVISION) otherwise nil."
...
+		    (if (bufferp file-or-buffer)
+			(buffer-file-name file-or-buffer)
+		      file-or-buffer)))
+	 (remote-url (git-link--remote-url remote))
+	 (remote-info (git-link--parse-remote remote-url))
+	 (branch      (git-link--branch))
+	 (commit      (git-link--commit))
+	 (handler     (git-link--handler git-link-remote-alist (car remote-info))))
+    (cond ((null remote-url)
+           (message "Remote `%s' not found" remote))
```
sshaw, line 687: Since `git-link-url` is called programmatically we should not call `(message)` as it will display an error in the mini-buffer. Instead, we should signal an error. See #45 for more info/examples.

git-link.el
```diff
@@ -688,43 +736,7 @@ Defaults to \"origin\"."
...
-                         (if (or (git-link--using-git-timemachine)
-                                 (git-link--using-magit-blob-mode)
-                                 vc-revison
-                                 git-link-use-commit)
-                             nil
-                           (url-hexify-string branch))
-                         commit
-                         start
-                         end))))))))
+  (git-link--new (git-link-url buffer-file-name remote start end)))
```
sshaw, line 739: Based on [my comment about `(message)`](https://github.com/sshaw/git-link/pull/69/files#r775096715) here will will now have catch any error and output it using `(message)`.

sshaw:
Hi, @futuro, lookin' good. Aside from my 2 comments I would squash this PR into a single commit. 

Merry Christmas! 🎄

git-link.el
futuro: I dig it, thank you for the suggestion!

futuro:
Hey @sshaw, thanks for the suggestions on `user-error`! I've never done error signaling/handling in elisp, so that was a fun excursion.

I've just pushed up that change, and it should be ready for review.

Just to clarify, are you asking me to squash all the commits in this PR, or are you saying that you're going to do a squash merge afterwards?

Thanks! And Happy New Year!

sshaw:
> Just to clarify, are you asking me to squash all the commits in this PR, or are you saying that you're going to do a squash merge afterwards?

Squash all commits, please.

futuro:
@sshaw commits are squashed and everything should be ready to go.

git-link.el
```diff
@@ -666,6 +666,54 @@ return (FILENAME . REVISION) otherwise nil."
...
+When given START and END, include the current line number or
+active region.
+
+When given USE-BRANCH-P, force the use of the branch name in the
+resultant url."
+  (let* ((filename (git-link--relative-filename
+                    (if (bufferp file-or-buffer)
+                        (buffer-file-name file-or-buffer)
+                      file-or-buffer)))
+         (remote-url (git-link--remote-url remote))
```
sshaw, line 681: We must check for a `nil` `remote-url` first, like we do here: https://github.com/sshaw/git-link/pull/69/files#diff-a86ea92b0e180064b421a637b78c7d5ae152de93c60a82136b77a6a72c90b35cL692
Otherwise we get errors when the remote does not exist.

git-link.el
```diff
@@ -688,43 +736,10 @@ Defaults to \"origin\"."
...
-                                 git-link-use-commit)
-                             nil
-                           (url-hexify-string branch))
-                         commit
-                         start
-                         end))))))))
+  (condition-case err
+      (git-link--new (git-link-url buffer-file-name remote start end))
+    (user-error
+     (message (cdr err)))))
```
sshaw, line 742: When an error occurs we get an error here: 

```
When linking in bufffer without git repo
condition-case: Wrong type argument: stringp, ("Can’t figure out what to link to")
condition-case: Wrong type argument: stringp, ("Can’t figure out what to link to")
```

We need `cadr` instead of `cdr`

git-link.el
```diff
@@ -666,6 +666,54 @@ return (FILENAME . REVISION) otherwise nil."
     (git-link--remote)))
 
 ;;;###autoload
+(defun git-link-url (file-or-buffer remote &optional start end use-branch-p)
```
sshaw, line 669: Some thoughts/issues here:

1. I think we should default `remote` to `git-link-default-remote`. 
1. If the current buffer is in directory `X` but this is called like `(git-link-url "/path/to/Y/foo.txt" "origin" 10)`, where `Y` is a different repository root, it should return a link to `Y/foo.txt`'s repo but now it links to something random in repo `X`. 
1. What to do with invalid line numbers, e.g., when `end` is after or equal to start, when `start` is `nil` and `end` is given? Thinking we check for these cases and signal an error.

sshaw (changes requested):
Hi, thanks for squashing. 

I had an in-depth look now and there are a few issues. See comments.

git-link.el
sshaw: A case where the remote does not exist would be trying to link from a buffer that is not part of a git repo.

git-link.el
futuro: Good catch, resolved.

git-link.el
sshaw: > ... but item 2 strikes me as out of scope for this PR.

When calling it as a separate function it cannot depend on state of some buffer. For example, if one creates an emacs package that calls `git-link-url` the call will depend on the state of some git repo outside the function call so bugs would occur. Make sense? 

> I agree with items 1 and 3, 

Maybe 2/3 is fine? Do they togeather outweigh the bug? 

> Do you see another way of achieving 2 that only impacts git-link-url?

I think this is part of the reason why this function has taken so long to exist. If I remember correctly you have to change a bit more than meets the eye. It has been since Dec or so since I had a look at the code. Will have to wait to look again. But if you think you gotz a solution then add it. But if you are concerned about your time then maybe wait 'till I can have an in-depth look again.

sshaw: Also test coverage not that good. [Exhibit A](https://github.com/sshaw/git-link/issues/95).

sshaw:
Closing to due lack of feedback/progress.

PR #72, opened 2020-07-02
Sourcegraph support

sideshowcoder:
Not sure this is something that would be an addition to git-link as sourcegraph does not actually implement a repository, but it is related to git.

Sourcegraph allows to reference files by git commit or branch, even so it does not represent a git repository it can be useful to open files in sourcegraph instead of the remote.

This requires a pseudo remote to be setup in the `.git/config` of the project.

```
[remote "sourcegraph"]
  url = https://my.sourcegraph.host/my.git.host/my-repository
[git-link]
  remote = sourcegraph
```

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -173,7 +173,8 @@
     ("bitbucket" git-link-bitbucket)
     ("gitorious" git-link-gitorious)
     ("gitlab" git-link-gitlab)
-    ("visualstudio\\|azure" git-link-azure))
+    ("visualstudio\\|azure" git-link-azure)
+    ("sourcegraph" git-link-sourcegraph))
   "Alist of host names and functions creating file links for those.
 Each element looks like (REGEXP FUNCTION) where REGEXP is used to
 match the remote's host name and FUNCTION is used to generate a link
@@ -190,7 +191,8 @@ As an example, \"gitlab\" will match with both \"gitlab.com\" and
     ("bitbucket" git-link-commit-bitbucket)
     ("gitorious" git-link-commit-gitorious)
     ("gitlab" git-link-commit-github)
-    ("visualstudio\\|azure" git-link-commit-azure))
+    ("visualstudio\\|azure" git-link-commit-azure)
+    ("sourcegraph" git-link-commit-sourcegraph))
   "Alist of host names and functions creating commit links for those.
 Each element looks like (REGEXP FUNCTION) where REGEXP is used to
 match the remote's host name and FUNCTION is used to generate a link
@@ -493,6 +495,27 @@ return (FILENAME . REVISION) otherwise nil."
 	  hostname
 	  dirname
 	  commit))
+
+(defun git-link-sourcegraph (hostname dirname filename branch commit start end)
+  (let ((line-or-range (cond ((and start end) (format "#L%s-%s" start end))
+                             (start (format "#L%s" start))
+                             (t "")))
+        (branch-or-commit (or branch commit))
+        (dir-file-name (directory-file-name dirname)))
+    (format "https://%s/%s@%s/-/blob/%s%s"
+            hostname
+            dir-file-name
+            branch-or-commit
+            filename
+            line-or-range)))
+
+(defun git-link-commit-sourcegraph (hostname dirname commit)
+  (let ((dir-file-name (directory-file-name dirname)))
+    (format "https://%s/%s/-/commit/%s"
+            hostname
+            dir-file-name
+            commit)))
+
 
 (defun git-link--select-remote ()
   (if current-prefix-arg


sshaw:
Hi, thanks, I've heard of this but have never used it –until now! I think it'd be a nice addition. 

Questions/comments: 

sourcegraph is only self-hosted, right? So alist value of "sourcegraph" assumes this is used by the sourcegraph configured domain?

Currently does not work with ports, may be an existing bug but sourcegraph runs on 7080 by default and generated links are missing the port 😿 

Locally I have a link like this: http://127.0.0.1:7080/github.com/ScreenStaring/recharge-api/-/blob/README.md?view=code#L10

But the resulting git-link is: 

http://127.0.0.1:7080/github.com/ScreenStaring/recharge-api/@master/-/blob/README.md#L10 (I added the port), which returns a 404

When `git-link-use-commit` is `t` I get: http://127.0.0.1:7080/github.com/ScreenStaring/recharge-api/@4bc47f20de32d1c182ffe63b12c94ce59057bd0c/-/blob/README.md#L10

Which also 404s. Any ideas? This is using the `sourcegraph/server:3.17.3` Docker image.

(Note that I also had to change it to http, but maybe that's not important now 😄)

sideshowcoder:
Thanks for considering @sshaw 

> sourcegraph is only self-hosted, right? So alist value of "sourcegraph" assumes this is used by the sourcegraph configured domain?

Yes I was under that assumption, but also if I understand the code right it depends on the naming of the remote, not the url in the remote am I correct? So with the configuration given above

```
[remote "sourcegraph"]
  url = https://i.dont.contain.the.magic.word/my.git.host/my-repository
[git-link]
  remote = sourcegraph
```

should also work, see the url not containing `sourcegraph`.

> Currently does not work with ports, may be an existing bug but sourcegraph runs on 7080 by default and generated links are missing the port 😿

Interesting I hadn't tried that, it seems to be an exisiting bug so, as the patch doesn't handle the parsing of the host, so I would assume that the port also fails for a self-hosted git running on a non default port for example. Happy to look into this, but I think its a seperate issue.

> Locally I have a link like this: http://127.0.0.1:7080/github.com/ScreenStaring/recharge-api/-/blob/README.md?view=code#L10
> 
> But the resulting git-link is:
> 
> http://127.0.0.1:7080/github.com/ScreenStaring/recharge-api/@master/-/blob/README.md#L10 (I added the port), which returns a 404
> 
> When `git-link-use-commit` is `t` I get: http://127.0.0.1:7080/github.com/ScreenStaring/recharge-api/@4bc47f20de32d1c182ffe63b12c94ce59057bd0c/-/blob/README.md#L10
> 
> Which also 404s. Any ideas? This is using the `sourcegraph/server:3.17.3` Docker image.

Interesting, I was running it against a company hosted sourcegraph instance, but I'll try the docker image and report back, I feel like there is something wrong with the branch / commit adding. I can't try it yet but I would assume the link http://127.0.0.1:7080/github.com/ScreenStaring/recharge-api@master/-/blob/README.md#L10 should work, there seems to be a rouge `/` at the end of the `dirname` part of the url, which would need to be dropped.

sideshowcoder:
@sshaw any chance you could take another look at this?

sshaw:
Oh, hi. Sorry, missed this.

> if I understand the code right it depends on the naming of the remote, not the url in the remote am I correct?

It does depend on the URL. It looks up the remote and tries to match its URL against the appropriate alist key.

> it seems to be an exisiting bug so, as the patch doesn't handle the parsing of the host... I think its a seperate issue.

Agreed. 

`git-link` is mostly working now (see inline comments). 

`git-link-commit` generates the following URL: http://127.0.0.1:7080/github.com/ScreenStaring/recharge-api@2cb1ad0 but for me this does not link to the commit. This links to the commit: http://127.0.0.1:7080/github.com/ScreenStaring/recharge-api/-/commit/84286a823bd1969366b5ff4abb22a2c3c5bf8ca4

Does the first one work for you? Does the second?

git-link.el
```diff
@@ -494,6 +496,25 @@ return (FILENAME . REVISION) otherwise nil."
...
+(defun git-link-sourcegraph (hostname dirname filename branch commit start end)
+    (let ((line-or-range (if end (format "%s-%s" start end) start))
+          (branch-or-commit (or branch commit))
+          (dir-file-name (directory-file-name dirname)))
+      (format "https://%s/%s@%s/-/blob/%s#L%s"
+              hostname
+              dir-file-name
+              branch-or-commit
+              filename
+              line-or-range)))
```
sshaw, line 508: `line-or-range` may be `nil`. For example, if you're linking from a magit-log or magit-status buffer. 

Currently it'll try to link to `#Lnil`

sshaw:
Also please squash into a single commit. 💪

git-link.el
sideshowcoder: Good point this should be fixed now.

sideshowcoder:
Hello @sshaw sorry about the delay, I finally got fix the issues and follow up. 

The link for `git-link-commit` was not working indeed but it seemed to be working on my instance as it did the correct redirect. It is working now.

sshaw:
No problem, thanks for this!

PR #74, opened 2020-08-23
Travis CI coverage for Emacs 27; Travis CI uses Nix

peterbecich:
Uses https://github.com/purcell/nix-emacs-ci now

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,64 @@
-# Mostly stolen from https://github.com/npostavs/emacs-travis/
-language: generic
-sudo: false
+# https://github.com/purcell/nix-emacs-ci
+language: nix
 
-env:
-  matrix:
-    - EMACS_VERSION=24.3
-    - EMACS_VERSION=24.5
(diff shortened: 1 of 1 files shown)


sshaw:
Ooooh Nix, nice...

peterbecich:
@sshaw , please review

sshaw:
@peterbecich thanks a lot. I recently updated Travis CI config for [rvm.el](https://github.com/senny/rvm.el) and saw that the Old Method ™️ didn't have the latest version so definitely will be using this there too!

Just one request: can you squash into a single commit?

Thanks! 💪

peterbecich:
@sshaw very good, done

sshaw:
One thing I just noticed is that [we explicitly call `make build`](https://github.com/sshaw/git-link/pull/74/files#diff-354f30a63fb0907d4ad57269548329e3R58). Since `make test` has `build` as a prerequisite, is there a reason to call it explicitly?

Merging regardless!

PR #79, opened 2021-03-04
[Savannah] Prevent extra "/git" appearing in M-x git-link

leungbk:
Prior to this patch, (cadr (git-link--parse-remote remote-url)) for
the Emacs repo would look like "git/emacs" instead of "emacs".

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -394,6 +394,13 @@ return (FILENAME . REVISION) otherwise nil."
                     "\\1/_git/"
                     path)))
 
+      ;; For Savannah
+      (when (string= "git.savannah.gnu.org" host)
+        (cond
+         ((string-match "\\`git/" path)
+          (setq path (substring path 4)))
+         ((string-match "\\`srv/git/" path)
+          (setq path (substring path 8)))))
 
       (list host path))))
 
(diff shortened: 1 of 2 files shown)


sshaw:
Hi, thanks. Give me a a day or three to study this. I mean it's straightforward but the `git-link--parse-remote` function is getting out of control with these exception cases so just want to take a bit to look it over.

git-link.el
```diff
@@ -394,6 +394,10 @@ return (FILENAME . REVISION) otherwise nil."
                     "\\1/_git/"
                     path)))
 
+      ;; For Savannah
+      (when (and (string-match "gnu\\.org\\'" host)
```
sshaw, line 398: Is there a benefit to matching the 2nd-level domain vs the complete domain? 

The downside which, I suppose is a valid use case is, it matches for something that's not savannah. Though with the path portion I suppose this does decrease the odds of this happening.

sshaw:
`--parse-remote` definitely needs some tweaking but that's a task for another day. 

The only comment I have besides the one I left inline is: while we're here shouldn't we add support for `ssh://git.savannah.gnu.org/srv/git/emacs.git` too?

git-link.el
leungbk: Fixed.

leungbk:
> The only comment I have besides the one I left inline is: while we're here shouldn't we add support for `ssh://git.savannah.gnu.org/srv/git/emacs.git` too?

Done.

sshaw:
Great thanks!

PR #84, opened 2021-05-04
Get filename from base-buffer when in an indirect buffer

zaeph:
Hi there,

This is a quick PR to fix the creation of links in indirect buffer.

I’ll use the opportunity to thank you for your work on this package. :)

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -330,7 +330,7 @@ return (FILENAME . REVISION) otherwise nil."
 (defvar magit-buffer-file-name)
 
 (defun git-link--relative-filename ()
-  (let* ((filename (buffer-file-name))
+  (let* ((filename (buffer-file-name (buffer-base-buffer)))
 	 (dir      (git-link--repo-root)))
 
     (when (null filename)


sshaw:
Oh an indirect buffer fix. Nice. What's your use case for indirect buffers? 

Thanks for the fix!

zaeph:
Thanks for the merge! 🎉

> What's your use case for indirect buffers?

It's particularly useful when you need to edit two places in parallel inside a single file.  For example, when I'm editing a module of a bigger function, I like to keep both in sight to periodically verify the outputs; cloning and narrowing makes this a breeze.

sshaw:
> ... when I'm editing a module of a bigger function, I like to keep both in sight to periodically verify the outputs; cloning and narrowing makes this a breeze.

Thanks. I'm going to try this. Currently I'd use multiple windows + [bm](https://github.com/joodland/bm) for bouncing around to/from important locations but what you said sounds good.

zaeph:
> Currently I'd use multiple windows + [bm](https://github.com/joodland/bm) for bouncing around to/from important locations but what you said sounds good.

That sounds interesting too; thanks for the discovery!

I'm working on a proof-of-concept patch for [abo-abo/lispy](abo-abo/lispy) that would facilitate my method.
Here's a demo (lisyp-ibuf-demo.mp4).

PR #87, opened 2021-08-04
adds service specific handlers support for git-link-homepage

wvandeun:
Adds support for service specific handlers for git-link-homepage.
I'm not familiar with all the services, for which there are already rules for other functions. Input on those would be appreciated.

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -235,8 +235,11 @@ As an example, \"gitlab\" will match with both \"gitlab.com\" and
   '(("git.sr.ht" git-link-homepage-github)
     ("github" git-link-homepage-github)
     ("bitbucket" git-link-homepage-bitbucket)
+    ("gitorious" git-link-homepage-github)
     ("gitlab" git-link-homepage-github)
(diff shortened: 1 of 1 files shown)


sshaw:
Thanks a lot. Much needed! 💪

> I'm not familiar with all the services, for which there are already rules for other functions. Input on those would be appreciated.

I say we just default to doing this if there's nothing in the alist:

```el
(format "https://%s/%s" base (cadr remote-info))
```

What do you think? 

It is same as the old behavior. Just having the alist is an improvement and if it doesn't work for some site(s) it can then be addressed at that time by... updating the alist.

wvandeun:
> Thanks a lot. Much needed! 💪
> 
> > I'm not familiar with all the services, for which there are already rules for other functions. Input on those would be appreciated.
> 
> I say we just default to doing this if there's nothing in the alist:
> 
> ```emacs-lisp
> (format "https://%s/%s" base (cadr remote-info))
> ```
> 
> What do you think?
> 

Done, used the git-link-homepage-github handler as the default for services that are not known to me.
This handler provides the desired behaviour.

sshaw:
Great. Excited to have this addition 🎉 

Thanks!

PR #88, opened 2021-08-09
Adding handler for codeberg.org

jeremyf:
codeberg.org uses the 'src' slug instead of 'blob' like Github.  This
commit helps produce the correct URL for Codeberg.


sshaw:
Hi, thanks for the PR. Never heard of code Codeberg but [I just signed up!](https://codeberg.org/sshaw) What's the story behind it?

That aside, since it looks like the commit and link functions are the same a GitHub, can we use its functions instead? E.g.

```el
 ("codeberg.org" git-link-github)
```
?

PR #90, opened 2021-10-27
Refactor into git-link-cgit & git-link-commit-cgit

hendursaga:
Savannah uses cgit for (one of) their web interfaces.

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -597,8 +597,8 @@ return (FILENAME . REVISION) otherwise nil."
 	  dirname
 	  commit))
 
-(defun git-link-savannah (hostname dirname filename branch commit start _end)
-  (format "https://%s/cgit/%s.git/tree/%s?h=%s"
+(defun git-link-cgit (hostname dirname filename branch commit start _end)
+  (format "https://%s/%s/tree/%s?h=%s"
 	  hostname
 	  dirname
           filename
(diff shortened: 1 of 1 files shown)


hendursaga:
Or were you meaning adding some tests, @sshaw?

sshaw:
Tests are always nice but I was wondering about public repos that I can use to test the linking. 

There [was talk of adding cgit support a while back](https://github.com/sshaw/git-link/pull/29#issuecomment-195653908), but the problem is the path portion is dependent on the cgit install. [cgit defaults to `cgit.cgi`](https://git.zx2c4.com/cgit/tree/Makefile#n4) but of course this can be anything. cgit's site does not (did not?) use `.cgi`, for example. 

How to workaround this issue? 

Working with SaaS services, e.g., GitHub makes sense but when the function is for software that anyone can configure, it doesn't seem to make sense to include in the package. Instead this is where a local, user-defined function in `git-link-remote-alist` and `git-link-commit-remote-alist` makes sense.

hendursaga:
That was why I added the optional base-path argument, although I honestly haven't seen any /cgit.cgi/ paths in the wild yet. Wouldn't this be sufficient?

sshaw:
> That was why I added the optional base-path argument

So if one needed to link to a repo with a cgit front-end that used `cgit.cgi` for the `SCRIPT_NAME` or, better yet `my-amaaaaazing-custom-url`, they'd have to do:

```el
(defun link-to-my-cgit-endpoint (hostname dirname filename branch commit start end)
  (git-link-cgit (hostname dirname filename branch commit start end "my-amaaaaazing-custom-url")))

;; add to alist stuff here

```
Correct?

> although I honestly haven't seen any /cgit.cgi/ paths in the wild yet

Org-Mode used to use it. It _is_ the default script name  🤷. Given that [it is the default](https://git.zx2c4.com/cgit/tree/Makefile#n4) (though I could be missing something. I have close to nil cgit experience), wouldn't it be best to default `base-path` to `cgit.cgi`?

hendursaga:
I realized a better way of doing this, vs using optional arguments - just modify the dirname variable a bit. Does this work better? It works out of the box for my other cgit-hosted repositories.

sshaw:
> It works out of the box for my other cgit-hosted repositories.

How did you install these because the question remains: if [the cgit Makefile defaults to `cgit.cgi`](https://git.zx2c4.com/cgit/tree/Makefile#n4), why don't you have `cgit.cgi`? 

Otherwise looks good. Just wondering what the disconnect is between that Makefile and everything else. Are folks just overriding it?

hendursaga:
> How did you install these...

I should clarify, when I said "my other cgit-hosted repositories," I don't mean to say I hosted or installed them, I just clone/push from/to them. I believe almost everyone installing `cgit` override that.

hendursaga:
@sshaw I believe I squashed it correctly! Ready to merge?

sshaw:
🎊 

Thanks again!

PR #92, opened 2022-01-27
URL escape special chars in pathnames

bitti:
I occasionally have to work on a project with whitespace in pathnames, which causes git-link to produce broken links. This change should fix that and other edge cases where problematic characters occur.


sshaw:
Great, thanks!

sshaw:
This is causing issues. See #93

PR #97, opened 2022-04-04
Add git-link-homepage-bitbucket

psibi:
Have been using this for a bunch of different bitbucket repos.


sshaw:
Hi, thanks for the PR. I didn't realize this was a bug as the function is referenced [here](https://github.com/sshaw/git-link/blob/f5691f8c66eb0f6050d9ab3834ab32661244b378/git-link.el#L254) but never defined (I'm using an older version before the homepage alist was added). 

Since this function is the same as GitHub's, can we update line 254 in the alist to refer to `git-link-homepage-github` instead of adding a new function?

psibi:
Thanks, done!

sshaw:
Great! Thanks again!

Issue #100, opened 2022-06-11
Feature request: support for Gitiles


peterbecich:
Request for a function `git-link-gitiles`

> Gitiles - A simple JGit repository browser
> Gitiles is a simple repository browser for Git repositories, built on JGit. Its guiding principle is simplicity: it has no formal access controls, no write access, no fancy Javascript, etc.

https://gerrit.googlesource.com/gitiles/

Example file in Gitiles: https://gerrit.googlesource.com/gitiles/+/refs/heads/master/java/com/google/gitiles/DateFormatter.java

sshaw:
Hi, thanks for the suggestion. 


Does the Google example represent the path that one would get with a default JGit installation? 

Also, if you can provide the following link structures for the follwing it would be helpful to explore adding this: 

1. a commit
2. a file with a line number
3. a file with a line number range (if supported)
4. a Markdown or similar file with a line number, i.e., an unrendered version of the Markdown file. GitHub uses `plain=1` for this.

peterbecich:
Thanks for the quick response

> Does the Google example represent the path that one would get with a default JGit installation?

I don't have any other examples of Gitile repositories outside of https://gerrit.googlesource.com/

I will guess that all Gitile repositories take this form:
```
[Git remote address]/+/[Git commit hash]/[file#line number]
```


For example repository https://gerrit.googlesource.com/bazlets/ with remote:
```
% git remote -v
origin  https://gerrit.googlesource.com/bazlets (fetch)
origin  https://gerrit.googlesource.com/bazlets (push)
```

1. An example commit: https://gerrit.googlesource.com/bazlets/+/efb3ad71f38e515eb1805ff457a077264b9f3dcc
2. A file with a line number: https://gerrit.googlesource.com/bazlets/+/efb3ad71f38e515eb1805ff457a077264b9f3dcc/lib/prolog/java/PrologCompiler.java#46
3. It appears a line number range is unsupported
4. It appears Markdown rendering can only be configured server-side: https://gerrit.googlesource.com/gitiles/+/HEAD/Documentation/config.md#disabling-markdown
There is no URL parameter I can find which disables it. Example markdown https://gerrit.googlesource.com/bazlets/+/efb3ad71f38e515eb1805ff457a077264b9f3dcc/README.md

peterbecich:
Here is another collection of Gitile repositories: https://git.eclipse.org/r/plugins/gitiles/?format=HTML

[...]

PR #105, opened 2022-12-28
Add support for googlesource.com

modulitos:
~~Resolves: https://github.com/sshaw/git-link/issues/100~~
(**Update:** this doesn't actually provide support for gitiles, and is limited to googelsource)

Also adds support for the repos hosted by `googlesource.com`, which I believe are all using gitiles. Such as https://android.googlesource.com/ and https://go.googlesource.com/go
I checked several projects under https://opensource.google/projects, and I believe all repos under the `googlesource.com` domain are using gitiles.

### Testing

I added a test to `git-link-test.el` to ensure the behavior of `git-link--parse-remote-test` was correct.

For the rest of the behavior, I ran the following tests manually, since I didn't see any existing automated tests (although it seems like the project can use some unit tests here):

#### `git-link`
I tested this by running `git-link` in several gitiles repos. 
The following value is copies to my clipboard:
1. with `git-link-use-commit` set to `t`: https://go.googlesource.com/go/+/38cfb3be9d486833456276777155980d1ec0823e/src/go/token/token.go#162
2. with `git-link-use-commit` set to `nil` (the default vaue): https://go.googlesource.com/go/+/master/src/go/token/token.go#162

#### `git-link-commit`
I tested this by running `git-link-commit` over some code like `// 38cfb3be9d486833456276777155980d1ec0823e` and got this:
https://go.googlesource.com/go/+/38cfb3be9d486833456276777155980d1ec0823e

#### `git-link-homepage`
Running `git-link-homepage` copies this to my clipboard: https://go.googlesource.com/go

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -214,6 +214,7 @@ See its docs."
     ("gitorious" git-link-gitorious)
     ("gitlab" git-link-gitlab)
     ("git\\.\\(sv\\|savannah\\)\\.gnu\\.org" git-link-savannah)
+    ("googlesource.com" git-link-googlesource)
     ("visualstudio\\|azure" git-link-azure)
     ("sourcegraph" git-link-sourcegraph)
     ("\\(amazonaws\\|amazon\\)\\.com" git-link-codecommit))
@@ -235,6 +236,7 @@ As an example, \"gitlab\" will match with both \"gitlab.com\" and
     ("gitorious" git-link-commit-gitorious)
     ("gitlab" git-link-commit-github)
     ("git\\.\\(sv\\|savannah\\)\\.gnu\\.org" git-link-commit-savannah)
+    ("googlesource.com" git-link-commit-googlesource)
     ("visualstudio\\|azure" git-link-commit-azure)
     ("sourcegraph" git-link-commit-sourcegraph)
     ("\\(amazonaws\\|amazon\\)\\.com" git-link-commit-codecommit))
@@ -255,6 +257,7 @@ As an example, \"gitlab\" will match with both \"gitlab.com\" and
     ("gitorious" git-link-homepage-github)
     ("gitlab" git-link-homepage-github)
     ("git\\.\\(sv\\|savannah\\)\\.gnu\\.org" git-link-homepage-savannah)
+    ("googlesource.com" git-link-homepage-github)
     ("visualstudio\\|azure" git-link-homepage-github)
     ("sourcegraph" git-link-homepage-github)
     ("\\(amazonaws\\|amazon\\)\\.com" git-link-homepage-codecommit))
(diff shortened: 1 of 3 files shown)


sshaw:
Hi @modulitos, thanks so much for your contribution and detailed explanation. Give me a week or so and I will review. Please follow-up if you don't hear back from me!

git-link.el
```diff
@@ -549,6 +556,16 @@ return (FILENAME . REVISION) otherwise nil."
                                 (format "L%s-L%s" start end)
                               (format "L%s" start)))))))
 
+(defun git-link-gitiles (hostname dirname filename branch commit start end)
+  (format "https://%s/%s/+/%s/%s"
```
sshaw, line 560: Shouldn't this be called `git-link-googlesource` as that is what the format matches?

[The problem with support for JGit and Gitiles](https://github.com/sshaw/git-link/issues/100#issuecomment-1159289486) is they are local severs and the URL format is not standard like it is for Google Source, GitHub, etc... One can configure the path to be anything they want, right?

git-link.el
```diff
@@ -35,6 +35,10 @@
 
 ;;; Change Log:
 
+;; 2022-12-27 - v0.8.7
+;; * Add support for [gitiles](https://gerrit.googlesource.com/gitiles) (Issue #100)
+;; * Add support for googlesource.com using gitiles
```
sshaw, line 40: This will get added when v0.8.7 —or whatever version— is made. No need to add it here.

git-link.el
```diff
@@ -35,6 +35,10 @@
 
 ;;; Change Log:
 
+;; 2022-12-27 - v0.8.7
+;; * Add support for [gitiles](https://gerrit.googlesource.com/gitiles) (Issue #100)
```
sshaw, line 39: Any reason for using Markdown here?

git-link.el
modulitos: Ah I see now - I believe this format only matches the Gitiles servers that are configured for google source. Sorry for the confusion :facepalm: 

Somehow I thought I had this figured out, but now I see that [Gitiles servers are configurable](https://gerrit.googlesource.com/gitiles/+/HEAD/Documentation/config.md#URLs).

I'll update this per your suggestion, and post a clarification in that thread as well :+1:

git-link.el
modulitos: Sounds good, I'll remove it :+1:

git-link.el
modulitos: No reason in particular, I thought the link would be helpful. I removed these comments anyway, per the thread above :+1:

modulitos:
@sshaw Thank you for the review! I made some updates, per your feedback. I think you're right, that this is limited to only googlesource.com repos :)

git-link.el
sshaw: I just don't know of anything that will render the Markdown so just extra syntax.

sshaw:
Great thanks. Can you just squash into single commit?

modulitos:
@sshaw Done :+1:

sshaw (approved):

PR #106, opened 2023-01-05
fix: empty branch name error, use commit hash as fallback always

liuyinz:
When you are not on any branch and set `git-link-use-commit` to `nil`, function `git-link--branch` return empty string "",  and returned url is wrong. So use commit hash as fallback if branch name is empty.

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -191,10 +191,12 @@
   :type 'boolean
   :group 'git-link)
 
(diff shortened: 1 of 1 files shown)


sshaw:
Hi, thanks for the PR. 

This has come up before in #46 but unfortunately we were unable to come to a conclusion. Have a look a the commentary there are let me know your thoughts.

PR #108, opened 2023-02-12
allow browser customizaton for git-link-open-in-browser

bb2020:
Hi. I've made a change so different `browse-url` functions can be used.
Example: `(setq git-link-open-in-browser 'eww-browse-url)`

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -187,8 +187,8 @@
   :group 'git-link)
 
 (defcustom git-link-open-in-browser nil
-  "If non-nil also open link in browser via `browse-url'."
-  :type 'boolean
+  "If non-nil also open link with this `browse-url-*' function."
+  :type '(choice boolean function)
   :group 'git-link)
 
 (defcustom git-link-use-commit nil
@@ -514,7 +514,9 @@ return (FILENAME . REVISION) otherwise nil."
   (message (replace-regexp-in-string "%" "%%" link t t))
   (setq deactivate-mark t)
   (when git-link-open-in-browser
-    (browse-url link)))
+    (if (fboundp git-link-open-in-browser)
+        (funcall git-link-open-in-browser link)
+      (browse-url link))))
 
 (defun git-link-codeberg (hostname dirname filename branch commit start end)
     (format "https://%s/%s/src/%s/%s"


git-link.el
```diff
@@ -514,7 +514,9 @@ return (FILENAME . REVISION) otherwise nil."
   (message (replace-regexp-in-string "%" "%%" link t t))
   (setq deactivate-mark t)
   (when git-link-open-in-browser
-    (browse-url link)))
+    (unless (fboundp git-link-open-in-browser)
+      (setq git-link-open-in-browser 'browse-url))
```
sshaw, line 518: Why set this, seems better to let it remain `t`? Is this because `defcustom`'s type changed?

git-link.el
```diff
@@ -187,8 +187,8 @@
   :group 'git-link)
 
 (defcustom git-link-open-in-browser nil
-  "If non-nil also open link in browser via `browse-url'."
-  :type 'boolean
+  "If non-nil also open link with this `browse-url-*' function."
+  :type 'function
```
sshaw, line 191: Possible to be `'function` and `'boolean`? Allowing `t` is nice for those that don't care.

sshaw: > Allowing t is nice for those that don't care.

How the URL is opened.

git-link.el
bb2020: This allows to be compatible with existing configs. If the value is `t` then `browse-url` will be used. Otherwise configured browser will be used.

git-link.el
bb2020: I presume `t` is still allowed. Because the value cell will be `t` and the function cell will be void. Don't quote me on that :smiley:

git-link.el
sshaw: Thanks. 

We can also do that without overwriting users' value: 

```el
(when git-link-open-in-browser
  (if (fboundp git-link-open-in-browser)
      (funcall git-link-open-in-browser link)
    (browse-url link)))
```

Not a fan of overwriting global values. Plus user may have some logic built on it that we then break.

bb2020: This is indeed better.
I thought user may see the change in `customize-group` and be informed about the new feature. It may not be necessary at all. That is why I overwrote its value.

bb2020: Should I change it?

bb2020: > Do people actually use Customize :)

I am using it, but I also use the laptop's DVD drive :rofl: 
I made the change. Please see my other reply about the `'function`.

git-link.el
bb2020: https://www.gnu.org/software/emacs/manual/html_node/elisp/Composite-Types.html
There is a choice type mentioned in the link. I don't think anyone has ever used it though :blush: 
`:type '(choice boolean function)`

sshaw: Okay, according to [the docs](https://www.gnu.org/software/emacs/manual/html_node/elisp/Composite-Types.html) we can do this: 

```el
:type '(list function boolean)
```

git-link.el
sshaw: > I am using it, but I also use the laptop's DVD drive 🤣

💽

git-link.el
bb2020: `list` means function and boolean are together in a list.

sshaw: Ah yes I see. `choice` sounds good.

sshaw (approved):
Thank you for this!

bb2020:
Awesome. Thanks for the fast review.

PR #109, opened 2023-02-15
Fix the Codeberg links to line ranges

vifon:


sshaw:
Thank you.

Did Codeberg change this or has it always been broken?

vifon:
It has been broken for as long as I'm aware, but that means only a few months back.

PR #112, opened 2023-05-17
gitlab: Use /-/ scoped routing

To1ne:
GitLab is changing toward having all project routes under the /-/ scope[1]. For a long time there was a redirect from legacy routes to these newer scoped routes, but in v16.0 this is no longer the case[2]. These scoped routes exist for several major versions already, so now it's safe to transition the GitLab URL to this newer scheme.

[1]: https://gitlab.com/gitlab-org/gitlab/-/issues/29572
[2]: https://gitlab.com/gitlab-org/gitlab/-/issues/28848


sshaw:
Thanks. 

For those that may be wondering why this is necessary see: https://gitlab.com/gitlab-org/gitlab/-/issues/214217

Issue #114, opened 2023-11-25
support for http


drdv:
Sometimes I need to have `http` instead of `https` in the links. Currently I deal with this using
```elisp
(use-package git-link
  :ensure t
  :bind ("C-c g l" . git-link)
  :config
  (setq git-link-open-in-browser t)
  (defun git-link--http-advice (func link)
    (funcall func (replace-regexp-in-string "^https:" "http:" link)))
  (defun git-link-toggle-http-advice ()
    "Toggle using http or https by advicing `git-link--new'."
    (interactive)
    (if (advice--p (advice--symbol-function 'git-link--new))
	(advice-remove 'git-link--new 'git-link--http-advice)
      (advice-add 'git-link--new :around 'git-link--http-advice)))
  (git-link-toggle-http-advice))
```
But clearly there is a better way to deal with this. It seems to me that it is trivial to support this in the code (e.g., by adding a variable instead of hard-coding `https` in the links).

sshaw:
What services do you need to link to using http?

drdv:
self-hosted gitlab

sshaw:
Fixed by #143

PR #115, opened 2023-11-25
add option to have http links

drdv:
Related to #114.

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -199,6 +199,11 @@ function set to that function's symbol."
   :type '(choice boolean function)
   :group 'git-link)
 
+(defcustom git-link-http-link nil
+  "If t use http instead of https in the link."
+  :type '(choice boolean function)
+  :group 'git-link)
+
 (defcustom git-link-use-commit nil
   "If non-nil use the latest commit's hash in the link instead of the branch name."
   :type 'boolean
@@ -285,6 +290,11 @@ Github, Gitlab, etc show a rendered version by default, for these extensions
 we can prevent that behaviour."
   :type 'list
   :group 'git-link)
+
+(defun git-link--link-type ()
+  (if git-link-http-link
+      "http"
+    "https"))
 
 (defun git-link--exec(&rest args)
   (ignore-errors
@@ -527,7 +537,8 @@ return (FILENAME . REVISION) otherwise nil."
       (browse-url link))))
 
 (defun git-link-codeberg (hostname dirname filename branch commit start end)
-    (format "https://%s/%s/src/%s/%s"
+    (format "%s://%s/%s/src/%s/%s"
+            (git-link--link-type)
 	    hostname
 	    dirname
 	    (or branch commit)
@@ -539,7 +550,8 @@ return (FILENAME . REVISION) otherwise nil."
                                 (format "L%s" start)))))))
 
 (defun git-link-gitlab (hostname dirname filename branch commit start end)
-  (format "https://%s/%s/-/blob/%s/%s"
+  (format "%s://%s/%s/-/blob/%s/%s"
+          (git-link--link-type)
 	  hostname
 	  dirname
 	  (or branch commit)
@@ -551,7 +563,8 @@ return (FILENAME . REVISION) otherwise nil."
                               (format "L%s" start)))))))
 
 (defun git-link-github (hostname dirname filename branch commit start end)
-  (format "https://%s/%s/blob/%s/%s"
+  (format "%s://%s/%s/blob/%s/%s"
+          (git-link--link-type)
 	  hostname
 	  dirname
 	  (or branch commit)
@@ -563,7 +576,8 @@ return (FILENAME . REVISION) otherwise nil."
                               (format "L%s" start)))))))
 
 (defun git-link-googlesource (hostname dirname filename branch commit start end)
-  (format "https://%s/%s/+/%s/%s"
+  (format "%s://%s/%s/+/%s/%s"
+          (git-link--link-type)
 	  hostname
 	  dirname
 	  (or branch commit)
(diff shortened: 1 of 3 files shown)


git-link.el
```diff
@@ -551,7 +563,8 @@ return (FILENAME . REVISION) otherwise nil."
                               (format "L%s" start)))))))
 
 (defun git-link-github (hostname dirname filename branch commit start end)
-  (format "https://%s/%s/blob/%s/%s"
+  (format "%s://%s/%s/blob/%s/%s"
```
sshaw, line 566: Hi, thanks for the PR. 

When would github ever use an `http` scheme?

You mentioned in #114 that you only use this for custom gitlab. For that case couldn't you use: 

```el
(eval-after-load 'git-link
 '(progn
     '("mygitlabhost git-link-gitlab))
```
❓ 

Certainly not opposed to this feature, but if there's not an `http` use case for the service, why add the code for the world to maintain? And if there is, what is upside to this over `git-link-remote-alist` addition above (Aside from being terser 😉)?

drdv: Indeed, my implementation is overly-simplistic and handles all hosts in the same way (regardless of the fact that the feature makes little sense when it comes to some of them). Point taken. 

To be concrete, your suggestion (as far as I understand it) would look like:
```el
(use-package git-link
  :config
  (eval-after-load 'git-link
    '(progn
       (add-to-list
	'git-link-remote-alist
	'("mygitlabhost"
	  (lambda (hostname dirname filename branch commit start end)
	    (format "http://%s/%s/-/blob/%s/%s"
		    hostname
		    dirname
		    (or branch commit)
		    (concat filename
			    (when start
			      (concat "#"
				      (if end
					  (format "L%s-%s" start end)
					(format "L%s" start))))))))))))
```
Sure, this achieves the desired effect, but I will have to maintain in my configuration a very explicit implementation of something that should be rather an internal detail (and I would have to make sure to keep it in sync as your library evolves). Furthermore, I would have to do a similar thing for `git-link-commit-gitlab` (quite verbose).

In this respect, my "advice" hack in #114 seems simpler to support (even though it relies on `git-link--new`). 

Let me know if you have another suggestion. If not, I can live with the current state of things (all this seems kind of a detail). 

Thanks for the support.

sshaw: >           (lambda (hostname dirname filename branch commit start end) (git-link-gitlab hostname dirname filename branch commit start end "http"))))

Slightly cleaner with `&rest` argument to lambda instead.

drdv: I agree with "... opening an issue X months later for this "bug" ..."

Your suggestions goes in the right direction. It boils down to:
```el
(use-package git-link
  :config
  (eval-after-load 'git-link
    (cl-loop
     for (alist . function) in '((git-link-remote-alist          . git-link-gitlab)
				 (git-link-commit-remote-alist   . git-link-commit-gitlab)
				 (git-link-homepage-remote-alist . git-link-homepage-github))
     do (add-to-list
	 alist
	 `("mygitlabhost"
	   (lambda (&rest args) (apply #',function (append args '("http")))))))))
``` 
This requires to add `&optional scheme` to the arguments of `git-link-gitlab`, `git-link-commit-gitlab` and `git-link-homepage-github` and in the `format` statement to use e.g., 
```el
(format "%s://%s/%s/-/blob/%s/%s"
    (or scheme "https")
    ...
```

Note that I had to use `git-link-homepage-github` because you reuse it in almost all entries in `git-link-homepage-remote-alist` so probably we have to define a dedicated function `git-link-homepage-gitlab` and I think that it would be good to rename `git-link-homepage-github` to something like `git-link-homepage-common`. 

I am not sure how using `git-link.scheme` would help. For me, having to add properties repository by repository to get a features like this is inconvenient. 

Having said all this, as a user, I would expect to have to do something like:
```el
(use-package git-link
  :config
  (git-link-register-gitlab-host "mygitlabhost" "http"))
```
where the library itself provides something like:
```el
(defun git-link-register-gitlab-host (host-name &optional scheme)
  (eval-after-load 'git-link
    (cl-loop
     for (alist . function) in '((git-link-remote-alist          . git-link-gitlab)
				 (git-link-commit-remote-alist   . git-link-commit-gitlab)
				 (git-link-homepage-remote-alist . git-link-homepage-github))
     do (add-to-list
	 alist
	 `(,host-name
	   (lambda (&rest args) (apply #',function (append args (when ,scheme '(,scheme))))))))))	   
```
but as an interface this looks strange (we handle one case - `gitlab`).

sshaw: > I am not sure how using git-link.scheme would help. For me, having to add properties repository by repository to get a features like this is inconvenient.

We want to avoid setting `git-link-http-link` for custom GitLab, causing http to get used for normal GitLab, GitHub, etc...  This is a per-repo setting. `git-link.scheme` is per-repo. 

On the elisp side, what's the repo/project version of a buffer-local variable? I don't think they exist.

> Having said all this, as a user, I would expect to have to do something like:

> (use-package git-link
>   :config
>   (git-link-register-gitlab-host "mygitlabhost" "http"))
  
Is this any more inconvenient than `git-link.scheme`? This package is 10 years old and this is the first time someone has needed it (at least publicly). http is the exception case. How often will one truly be inconvenienced by this?

drdv: > On the elisp side, what's the repo/project version of a buffer-local variable? I don't think they exist.

What do you mean?

> We want to avoid setting git-link-http-link for custom GitLab, causing http to get used for normal GitLab, GitHub, etc... This is a per-repo setting. git-link.scheme is per-repo.

I don't insist about any of this - I just share my opinion. For me the most convenient way to resolve this is to define "properties" per host (and not per project). Doing something per project wouldn't fit my workflow (but this is only me, I guess).

 > This package is 10 years old and this is the first time someone has needed it

I can live with my own hack (there is no problem).

sshaw: >> On the elisp side, what's the repo/project version of a buffer-local variable? I don't think they exist.

> What do you mean?

The http/https setting is per project, right? In the current PR setting `git-link-http-link` will set it globally. As far as I know there's no way to do this via Emacs Lisp using a variable. There's buffer-local which kinda works but buffer scope is way too narrow. 

> I don't insist about any of this - I just share my opinion. For me the most convenient way to resolve this is to define "properties" per host (and not per project). Doing something per project wouldn't fit my workflow (but this is only me, I guess).

To me it seems very simple for current PR to replace `git-link-http-link`  check with `git-link.scheme` git config check. The property approach is broader than that, and may even be functionality on top of this. Let me take a look at your config a bit more to make sure I understand.

PR #116, opened 2023-12-08
Ability to resolve using OpenSSH client configuration file

psibi:
So I have this setup where I use a different github account and a different private key configured like this in my ~/.ssh/config

```
Host github+myorg
  IdentitiesOnly yes
  HostName github.com
  IdentityFile ~/.ssh/myorg_id_rsa
  ```

This PR provides the ability to support such configuration. I have disabled it by default since this is not a common setup most peoples have.

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -214,6 +214,11 @@ See its docs."
   :type 'boolean
   :group 'git-link)
 
+(defcustom git-link-consider-ssh-config nil
+  "Consider ssh configuration file for resolving hostnames."
+  :type 'boolean
+  :group 'git-link)
+
 (defcustom git-link-remote-alist
   '(("git.sr.ht" git-link-sourcehut)
     ("codeberg.org" git-link-codeberg)
@@ -431,9 +436,15 @@ return (FILENAME . REVISION) otherwise nil."
       (when (string-match ":" host)
         (let ((parts (split-string host ":" t))
               (case-fold-search t))
-          (string-match (concat (car parts) ":\\(" (cadr parts) "\\)/") url)
+          (string-match (concat (regexp-quote (car parts)) ":\\(" (cadr parts) "\\)/") url)
           (setq host (car parts)
                 path (concat (match-string 1 url) "/" path))))
+
+
+      (when git-link-consider-ssh-config
+	(let* ((ssh-resolved-host (git-link--ssh-resolve-hostname host)))
+	  (when ssh-resolved-host
+	    (setq host ssh-resolved-host))))
 
       ;; Fix-up Azure SSH URLs
       (when (string= "ssh.dev.azure.com" host)
@@ -470,6 +481,15 @@ return (FILENAME . REVISION) otherwise nil."
                              (substring path 9)))))
 
       (list host path))))
+
+(defun git-link--ssh-resolve-hostname (hostname)
+  "Resolve HOSTNAME using ssh client."
+  (let ((output (shell-command-to-string (format "ssh -G %s" hostname)))
+	(host nil))
+    (dolist (line (split-string output "\n"))
+      (when (string-match "^hostname \\(.*\\)" line)
+	(setq host (match-string 1 line))))
+    host))
 
 (defun git-link--using-git-timemachine ()
   (and (boundp 'git-timemachine-revision)

### CODE REVIEW

**Comment by sshaw** [comment]
Hi, thanks for the PR!

To make sure I understand, your git remote uses `github+myorg` but you want that to link to `github.com`, is that correct?

File: git-link.el [hunk 1]
```diff
@@ -431,10 +436,16 @@ return (FILENAME . REVISION) otherwise nil."
       (when (string-match ":" host)
         (let ((parts (split-string host ":" t))
               (case-fold-search t))
-          (string-match (concat (car parts) ":\\(" (cadr parts) "\\)/") url)
+          (string-match (concat (regexp-quote (car parts)) ":\\(" (cadr parts) "\\)/") url)
```
sshaw, line 439: 👍

git-link.el
```diff
@@ -471,6 +482,30 @@ return (FILENAME . REVISION) otherwise nil."
...
+
+(defun git-link--parse-ssh-config ()
+  "Parse ~/.ssh/config and return a list of Struct with host and hostname."
+  (let ((current-host nil)
+	(resolved-hostname nil)
+        (hosts nil))
+    (with-temp-buffer
+      (insert-file-contents "~/.ssh/config")
+      (dolist (line (split-string (buffer-string) "\n"))
+	(cond ((string-match "^Host[ ]\\(.*\\)" line)
```
sshaw, line 495: I wonder if this can be done with `tramp-get-connection-property`? I had a quick looksy and the first argument was a bit confusing so, maybe not?

psibi: Yes, that function won't help. Infact, I initially looked into tramp to see if I can re-use any functions from there.

The closest I could find was `tramp-parse-sconfig`, but unfortunately that doesn't give information about the resolved hostname.

psibi:
> To make sure I understand, your git remote uses github+myorg but you want that to link to github.com, is that correct?

Yes. The remote for my repository looks like this:

```
git@github+myorg:MyOrganization/repo-name.git
```

git-link.el
sshaw: Giving that an ssh config `Host can contain wildcards, thinking it makes sense to pass it through `ssh -G`. It is a bit of an edge case, but one upside is parsing its output seems easier. Thoughts?

psibi: Yeah, I think that would be probably better. I will update this PR once I get some spare cycles.

psibi: Updated implementation to use `ssh -G`.

git-link.el
```diff
@@ -471,6 +482,15 @@ return (FILENAME . REVISION) otherwise nil."
 
       (list host path))))
 
+(defun git-link--ssh-resolve-hostname (hostname)
+  "Resolve HOSTNAME using ssh client."
+  (let ((output (shell-command-to-string (format "ssh -G %s" hostname)))
+	(host nil))
+    (dolist (line (string-split output "\n"))
```
sshaw, line 489: `string-split`, where does this come from? I do not have it in my Emacs. Using `27.2`

sshaw: Also on Window will this be `\n` or `\r\n`?

psibi: It comes from subr.el which comes with Emacs 29.1. I will see what I can do to support older versions.

sshaw: What about this? 

```el
(defun find-it (host)
  (with-temp-buffer
    (call-process "/usr/bin/ssh" nil (list (current-buffer) nil) nil "-G" host)
    (beginning-of-buffer)
    (when (search-forward-regexp "^hostname \\(.*\\)")
      (match-string 1))))
```
Should take care of the Windows line terminator problem. If it was even a problem.

sshaw: Well `(defun git-link--ssh-resolve-hostname ...)` 😉

sshaw: Should also just be `"ssh"`, right?

psibi: @sshaw Sorry, this took more time to come back than what I expected. I have replaced `string-split` with `split-string` which is available from Emacs 20.1. In fact `string-split` is a `defalias` to it.

git-link.el
```diff
@@ -471,6 +482,15 @@ return (FILENAME . REVISION) otherwise nil."
 
       (list host path))))
 
+(defun git-link--ssh-resolve-hostname (hostname)
+  "Resolve HOSTNAME using ssh client."
+  (let ((output (shell-command-to-string (format "ssh -G %s" hostname)))
+	(host nil))
+    (dolist (line (split-string output "\n"))
+      (when (cond ((string-match "^hostname \\(.*\\)" line) (match-string 1 line)))
```
sshaw, line 490: Do we need `cond`? If `(string-match "^hostname \\(.*\\)" line)` returns `t` then we know `(match-string 1 line)` will be non-`nil`, right?

git-link.el
sshaw: Hi, great thanks for the update!

Do we know `\n` will work on Windows? It's okay if not. 

Can you squash into a single commit?

Also added one question about `cond`

git-link.el
psibi: You're right, it's not needed. I have simplified it based on your feedback. Thanks!

git-link.el
psibi: > Do we know \n will work on Windows? It's okay if not.

It won't work. I can make it portable if you want, but I wouldn't be able to test it on a Windows machine since I don't have one.

> Can you squash into a single commit?

Done, thanks!

sshaw: >> Do we know \n will work on Windows? It's okay if not.

> It won't work. I can make it portable if you want, but I wouldn't be able to test it on a Windows machine since I don't have one.

And I just uninstalled my Virtual Box Windows (7!) image. Should just be a matter of changing `"\n"` to `"\r?\n"` but we can do that latter. 

Thanks for the work!

sshaw:
Release in v0.9.1

psibi:
Thank you!

PR #125, opened 2024-06-19
git-link: Optional opening in browser

psibi:


git-link.el
```diff
@@ -893,15 +893,19 @@ Defaults to \"origin\"."
...
 
 With a prefix argument prompt for the remote's name.
 Defaults to \"origin\"."
 
   (interactive (list (git-link--select-remote)))
-  (let* (handler remote-info (remote-url (git-link--remote-url remote)))
+
+  (let* (handler remote-info
+		 (remote-url (git-link--remote-url remote))
+		 (git-link-open-in-browser (or git-link-open-in-browser (equal 1 current-prefix-arg))))
```
sshaw, line 907: Instead of a prefix arg of 1 what do you think about using a double-prefix (<kbd>Ctrl-u Ctrl-u</kbd>) since that's what `git-link` uses? Makes it more consistent across functions. 

Seems that single prefix is what we want but that prompts for remote which I _guess_ is necessary to keep inline with `git-link` since homepage could be based on remote name.

psibi: Sure, modified to use double-prefix instead.

sshaw:
Great thanks!

PR #126, opened 2024-07-12
add gitea support

### LINKED ISSUE #117: support 'Gitea' 's  (and its fokr : Forgejo ?)  server ?
Hi
  I run my own git ( ['Gitea' server](https://gitea.com/) ) ... which is becoming a more and more popular 'self-hosted' option !
that would be great to enjoy git-link there too

thX thX thX

port19x:
closes #117

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,7 @@ Name of the remote branch to link to.
 * [cgit](https://wiki.archlinux.org/title/Cgit)
 * [GitHub](http://github.com)
 * [GitLab](https://gitlab.com)
+* [Gitea](https://about.gitea.com/)
 * [Gitorious](http://gitorious.org)
 * [GoogleSource](https://googlesource.com)
 * [Savannah](https://git.savannah.gnu.org/cgit)
--- a/git-link.el
+++ b/git-link.el
@@ -243,7 +243,8 @@ See its docs."
     ("googlesource.com" git-link-googlesource)
     ("visualstudio\\|azure" git-link-azure)
     ("sourcegraph" git-link-sourcegraph)
-    ("\\(amazonaws\\|amazon\\)\\.com" git-link-codecommit))
+    ("\\(amazonaws\\|amazon\\)\\.com" git-link-codecommit)
+    (".*\\|git\\|.*" git-link-codeberg))
   "Alist of host names and functions creating file links for those.
 Each element looks like (REGEXP FUNCTION) where REGEXP is used to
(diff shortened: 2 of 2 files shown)


port19x:
my very broad regex will probably break some customizations[¹](https://github.com/sshaw/git-link/tree/master?tab=readme-ov-file#building-links-and-adding-services)
This is because alist-get fetches the first match and custom domains will likely match my regex before reaching the appended alist entry

sshaw:
Hi, thanks for the PR. 

> my very broad regex will probably break some customizations[¹](https://github.com/sshaw/git-link/tree/master?tab=readme-ov-file#building-links-and-adding-services)

Is the use of Gitea so widespread that it is worth breaking these customizations? 

Since Gitea is self-hosted and has all sorts of URLs is it even needed in list library [when one can cal `add-to-list`as needed](https://github.com/sshaw/git-link?tab=readme-ov-file#building-links-and-adding-services) using the appropriate config-specific URL?

port19x:
It sees a fair amount of use by corporates and self-hosting types.
A way to integrate it that does not break these customizations would be to modify the `alist-get` calls.
Adding the right appropiate backends as the third argument would set it as default, which gives further opportunity to downside the alists, especially since commit and homepage are fairly uniform.

sshaw:
> A way to integrate it that does not break these customizations would be to modify the alist-get calls.
> Adding the right appropiate backends as the third argument would set it as default, which gives further opportunity to 
> downside the alists, especially since homepage is fairly uniform.

This seems like unnecessary work given the alternative: 

```el
(eval-after-load 'git-link
 '(progn
     '("your-custom-gitea-url" git-link-codeburg))
```

Maybe a `defalias` is needed so that one can say `git-link-gitea` instead? 

This situation is similar to Cgit which we currently support via a custom function but it is not added to `git-link-remote-alist` etc... because there is no definitive URL like GitLab, etc...

sshaw:
> This seems like unnecessary work given the alternative:
> (eval-after-load 'git-link
>   '(progn
>     (add-to-list 'git-link-remote-alist
> ...

Now admittedly having to add to **3** lists is not ideal and should change. I think #115 has a proposed fix for this

port19x:
Well then, given the workaround, I'll close the PR.
Anyone curious enough to check the issue tracker for support will find what backends to set, so that's fine.

sshaw:
Updated docs with info on how to enable Gitea support and added `git-link-*gitea` functions

PR #127, opened 2024-07-30
Only hexify branch name if it's not nil.

mijoharas:
When we pass the branch name through to the handler, we do this https://github.com/sshaw/git-link/blob/master/git-link.el#L864
that calls `url-hexify-string` on the branch, whether it's nil or not. Passing a `nil` branch through to it will give `""` which is truthy, which means that the `(or branch commit)` line in each of the handlers, e.g.: https://github.com/sshaw/git-link/blob/master/git-link.el#L594 will always evaluate to the truth-ey empty string `""` and will give an incorrect url.

Fix that by only calling `url-hexify-string` if the branch is non-nil already.

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -861,7 +861,9 @@ With a double prefix argument invert the value of
                                      (not git-link-use-commit)
                                    git-link-use-commit))
                              nil
-                           (url-hexify-string branch))
+                           (if branch
+                               (url-hexify-string branch)
+                             nil))
                          commit
                          start
                          end))))))))


sshaw:
Thanks.

How can I reproduce this issue? 

Even though we don't call `url-hexify-string` the format string `"%s"` still exists so won't that get interpolated with `nil` anyways?

mijoharas:
hey, so, I just called `git-link` on a commit that wasn't part of a branch. i.e. from this commit, we do `git co HEAD~1`:

and then I get something like https://github.com/sshaw/git-link/blob//README.md?plain=1#L1 (which is a link that doesn't exist) instead of the expected https://github.com/sshaw/git-link/blob/0acdfd6152fad132cecf25f6971c7f6b430dca15/README.md?plain=1#L1
(i.e. the empty string gets interpolated, but that leads to an invalid url).

sshaw:
Oh yes, I see. Thank you!

mijoharas:
ahhh, sorry should have checked the issues, could have linked to them myself :sweat_smile: (went straight to the code to just figure out what was wrong). Thanks again :raised_hands:

Issue #83, opened 2021-03-23
Support for Linking to `git blame` Page & Linking to Documentation Files like Org and Markdown Files on Bitbucket


djgoku:
Whenever using `git-link` on these files for GitHub/Bitbucket I usually have to use the blame/annotate url to link vs GitLab does the what I would expect.

https://gitlab.com/protesilaos/modus-themes/-/blob/main/README.md#L1-13 vs https://gitlab.com/protesilaos/modus-themes/-/blob/main/README.md

vs 

https://github.com/sshaw/git-link/blob/master/README.md#L6-L11 (does nothing, other than open the link) vs https://github.com/sshaw/git-link/blame/master/README.md#L6-L11 links to and highlights lines 6-11.

vs

https://bitbucket.org/tutorials/markdowndemo/src/master/README.md?at=master#README.md-1:11 (does nothing, other than open the link) vs https://bitbucket.org/tutorials/markdowndemo/annotate/master/README.md?at=master#README.md-1:11

If I was wanting to add this feature where would be the best place to add this?

sshaw:
Hi, 

It sounds like there are a few things here. First bugs, for Org/Markdown files:

* GitHub: linking to line numbers in does not work. This appears to be a GitHub bug because if you do it by hand the URL is the same.
* Bitbucket: no longer support linking to line numbers (brilliant Atlassian) unless going through the blame page

Second, a new feature:

* Linking to the blame page for a file which you're saying can be (or you want to be) a solution for the 2 items above

Is that correct?

djgoku:
> * GitHub: linking to line numbers in does not work. This appears to be a GitHub bug because if you do it by hand the URL is the same.
> * Bitbucket: no longer support linking to line numbers (brilliant Atlassian) unless going through the blame page

Yes, if I `git-link` on either an `org-mode` file or a Markdown file for GitHub/Bitbucket when opening the link it'll just open the file. If I use blame/annotate I can link to specific line numbers. This is a GitHub/Bitbucket specific issue.

> * Linking to the blame page for a file which you're saying can be (or you want to be) a solution for the 2 items above

Yes I want to write something to handle the above situations so I don't have to manually create the links.

sshaw:
Fixed by #128

djgoku:
@sshaw did you mean to re-open this?

sshaw:
No. Strange.

PR #128, opened 2024-09-22
Chore/add bitbucket tests

djgoku:
This implements #83 to enhance the generation of bitbucket.org links for markdown files.

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -304,6 +304,15 @@ we can prevent that behaviour."
   :type 'list
   :group 'git-link)
 
+;; https://support.atlassian.com/bitbucket-cloud/docs/readme-content/#Extensions-and-Languages
+(defcustom git-link-extensions-rendered-via-bitbucket-annotate '("org" "md" "mkd" "mkdn" "mdown"
+                                                                 "markdown" "text" "rst" "textile"
+                                                                 "asciidoc")
+  "List of extensions that should be rendered via annotate else
+they will actually be rendered. We can prevent that behaviour."
+  :type 'list
+  :group 'git-link)
+
 (defun git-link--exec(&rest args)
   (ignore-errors
     (with-temp-buffer
@@ -678,9 +687,10 @@ return (FILENAME . REVISION) otherwise nil."
 
 (defun git-link-bitbucket (hostname dirname filename _branch commit start end)
   ;; ?at=branch-name
-  (format "https://%s/%s/src/%s/%s"
+  (format "https://%s/%s/%s/%s/%s"
           hostname
           dirname
+          (git-link--should-render-via-bitbucket-annotate filename)
           commit
           (if (string= "" (file-name-nondirectory filename))
               filename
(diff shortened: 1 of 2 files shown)


sshaw:
Hi, thanks for the PR. Glad to see this was still on your radar! 📡

git-link.el
```diff
@@ -678,9 +678,12 @@ return (FILENAME . REVISION) otherwise nil."
 
 (defun git-link-bitbucket (hostname dirname filename _branch commit start end)
   ;; ?at=branch-name
-  (format "https://%s/%s/src/%s/%s"
+  (format "https://%s/%s/%s/%s/%s"
           hostname
           dirname
+          (if (string= "md" (downcase (or (file-name-extension filename) "")))
```
sshaw, line 684: Markdown also supports the `.markdown` extension. Given this, and the fact that the problem can occur with other markup formats, should we introduce a list, something like `bitbucket-non-source-formats` then throw `org` in there too?

djgoku: Ha, I knew you were going to suggest a list. Also interesting about org files. Here is an example org file image and this is the url for this: https://bitbucket.org/johnny-5-is-alive/dot-files/src/8b36a0e425ef7a002d8de0ff57a3ec81e39b075e/dev-setup-arch.org#lines-18:19

This is all to say I think for now we can support `.md` and `.markdown`. There isn't any harm in adding `org`. Let me know if I should just add `org` too.

sshaw: > This is all to say I think for now we can support .md and .markdown. There isn't any harm in adding org. Let me know if I should just add org too.

No need to add Org if they don't render it by default. 

The formats that they do render may be here: https://support.atlassian.com/bitbucket-cloud/docs/readme-content/, though unless you're looking for more to do just `.md` and `.markdown` are fine for now. 

Looking at [their docs](https://support.atlassian.com/bitbucket-cloud/docs/hyperlink-to-source-code-in-bitbucket/) it _sounds_ like the `fileviewer` query string parameter could change the view but I couldn't not find nor guess what that value would be. Similar questions [here](https://stackoverflow.com/search?q=fileviewer+%5Bbitbucket%5D). 

> @sshaw or do you just want me to reuse git-link-extensions-rendered-plain

No need. It provides for different functionality.

sshaw: > Looking at [their docs](https://support.atlassian.com/bitbucket-cloud/docs/hyperlink-to-source-code-in-bitbucket/) it sounds like the fileviewer query string parameter could change the view but I couldn't not find nor guess what that value would be. Similar questions [here](https://stackoverflow.com/search?q=fileviewer+%5Bbitbucket%5D).

Probably a good question to ask at https://community.atlassian.com/.

While we're talkin' BitBucket, their source view with that tiny `div` that stops short of the screen height and forces one to scroll is total garbage. What were they thinking!

djgoku: @sshaw I went the route you suggested.

git-link.el
```diff
@@ -304,6 +304,15 @@ we can prevent that behaviour."
   :type 'list
   :group 'git-link)
 
+;; https://support.atlassian.com/bitbucket-cloud/docs/readme-content/#Extensions-and-Languages
+(defcustom git-link-extensions-rendered-via-annotate-bitbucket '("org" "md" "mkd" "mkdn" "mdown"
```
sshaw, line 308: Sounds a wee bit better if we swap `bitbucket` and `annotate`: `git-link-extensions-rendered-via-bitbucket-annotate`

sshaw:
Great, thanks!

Issue #129, opened 2024-10-22
Transient integration is breaking doomemacs setup


vlnn:
After upgrading to latest version of git-link doom emacs starts to fail during the start. 

```
Warning (initialization): An error occurred while booting Doom Emacs:

Symbol's function definition is void: transient-prefix
To ensure normal operation, you should investigate and remove the
cause of the error in your Doom config files. Start Emacs with
the '--debug-init' option to view a complete error backtrace.
Error (doom-after-init-hook): Error running hook "doom-display-benchmark-h"
because: (wrong-type-argument hash-table-p nil)

```

It looks like the issue below somehow got to the git-link, as after disabling it in the packages emacs is accessible again.

https://github.com/magit/transient/issues/254

jplindstrom:
One more data point: Happens with `use-package` with `straight` too.

dpassen:
`use-package` with `elpaca` also fails to install the latest.

sshaw:
@blahgeek ideas on how to fix? 

Reverted the commits for now. It seems no ones uses MELPA stable?

sshaw:
> Reverted the commits for now. It seems no ones uses MELPA stable?

Well other options too. 

How are you pulling in this package @dpassen @jplindstrom @vlnn @vespesa❓

necaris:
Note: using the version without the Transient integration seems to work for me -- using Doom Emacs, the following line in my `packages.el` pins the commit:
```elisp
  (package! git-link :pin "fd33c3f7692bdf4b039afaa88cf07580429043ad")
```
It seems that for compatibility with Emacs 29.x you need a [slightly different autoload declaration for Transient](https://github.com/magit/transient/issues/280#issuecomment-2019214088)?

vlnn:
Revert (and moving transient functionality to another repo) helped in my case, so pinning to previous commits should not be needed anymore. Thanks @sshaw!

PR #132, opened 2024-11-22
Re-add transient menu

deejayem:
https://github.com/sshaw/git-link/pull/121 added a transient menu, but this did not work with any currently released versions of Emacs (https://github.com/sshaw/git-link/issues/129), so it was reverted.

I cherry-picked the two original commits, and added a third commit that fixes the issue (based on https://github.com/magit/transient/issues/280#issuecomment-2019214088) by adding `(autoload 'git-link-dispatch "git-link-transient" nil t)` to the autoload line. I cherry-picked the commits as I didn't want to take credit for anyone else's work, but let me know if you want it done differently.

This fixes the issue for me on 29.4.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,14 @@ This defaults to `"sourcegraph"` but can be changed. See [Building Links and Add
 
 URLs with ports or an http scheme will not work. It's a trivial fix so if it's a problem for you please open an issue.
 
+### [Emacs Transient](https://www.gnu.org/software/emacs/manual/html_mono/transient.html) Support
+
+An optional Transient interface (magit-like menu) is provided via `git-link-transient.el`. To enable you need to have
+`transient` installed as a dependency.
+
+To enable `(require 'git-link-transient)` and call `git-link-dispatch` to show the menu.
+
+
 ### Building Links and Adding Services
 
 `git-link-remote-alist` is an alist containing `(REGEXP FUNCTION)`
--- a/git-link-transient.el
+++ b/git-link-transient.el
@@ -1,0 +1,128 @@
+;;; git-link-transient.el --- Transient interface for git-link  -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2013-2024 Skye Shaw and others
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; Transient interface (magit-like menu) for git-link.
+;; Call `git-link-dispatch' to show the menu.
+;;
+;; You need to have `transient' installed as a depenency.
+;; (it's not listed as the dependency of git-link because we want it to be optional.)
+
+;;; Code:
+
+(require 'cl-lib)
+(require 'transient)
+(require 'git-link)
+
+(defun git-link-dispatch--action ()
+  "Finally call `git-link' with transient arguments."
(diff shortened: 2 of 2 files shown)


sshaw:
Hi, thanks for this. 

I didn't study #129 too closely but I assumed `git-link-transient.el` was somehow being loaded and `transient` was not defined but after seeing this PR looks like that is not `t`rue. 

How did you reproduce the issue in #121 and have you tried this on versions of emacs that do not have `transient`?

deejayem:
> Hi, thanks for this.
> 

Thank you for creating git-link! I typically use it multiple times per week, and it saves me a lot of time.

> I didn't study #129 too closely but I assumed `git-link-transient.el` was somehow being loaded and `transient` was not defined but after seeing this PR looks like that is not `t`rue.
> 
> How did you reproduce the issue in #121 and have you tried this on versions of emacs that do not have `transient`?

I was able to reproduce the issue on 29.x by installing the original version of git-link-transient.el and starting Emacs, giving this error:
`⛔ Warning (emacs): Error loading "git-link" autoloads: (void-function transient-prefix)`
After the change in my extra commit, 29.x starts without error, and `git-link-dispatch` works.

On 27.x, I see the same behaviour in terms of the error at startup. The difference is that if transient isn't installed as a dependency of something else, when I run `git-link-dispatch` I get the error `Cannot open load file: No such file or directory, transient`. (Transient was added to Emacs in 28.1).

Based on this comment in `git-link-transient.el` I think that's the intended behaviour:
```
;; You need to have `transient' installed as a dependency.
;; (it's not listed as the dependency of git-link because we want it to be optional.)
```

I think it would be better to depend on transient rather than it being broken, but it's up to you. Anyone who has magit installed will already have transient, as will anyone on 28.1+.

(By the way, unless you're using something like straight or elpaca, the easiest way to test it is to copy `git-link-transient.el` into the git-link directory in `~/.emacs.d/elpa/` and then run `(package-generate-autoloads "git-link" "~/.emacs.d/elpa/git-link-20241022.2208/")`).

deejayem:
@sshaw Do you have any thoughts on how you would like to handle this?

I can think of three possibilities:

1) Use the approach already taken on this PR. This means that with 28.1+, everything will work fine. For older versions, there will be no errors at startup (unlike with the original version), but `git-link-dispatch` will give an error unless transient is already installed. (I'm not keen on that, but if people also use magit they won't have an issue).

2) Add transient to `Package-Requires`. This would require increasing the minimum supported Emacs version to 26.1.

3) Create a separate git-link-transient package. Either you or @blahgeek could do that, but I can do it if neither of you want to.

There may be some other approach that I haven't thought of though.

sshaw:
Hi, 

Thanks for all the info!

> 1. Use the approach already taken on this PR. This means that with 28.1+, everything will work fine. For older versions, there will be no errors at startup (unlike with the original version), but git-link-dispatch will give an error unless transient is already installed. (I'm not keen on that, but if people also use magit they won't have an issue).

`git-link-dispatch` would only be called if they wanted to use the transient menu, right? If so I think this is okay and is sorta typical Emacs behavior. As long as non git-link transient users are not affected. 
 
> Add transient to Package-Requires. This would require increasing the minimum supported Emacs version to 26.1.

I try to keep dependencies at a minimum though supporting 26 in 2024/25 does not seem too crazy. But I think point 1 is better approach. Though why 26 if transient was adding in 28?

deejayem:
> Hi,
> 
> Thanks for all the info!
> 
> > 1. Use the approach already taken on this PR. This means that with 28.1+, everything will work fine. For older versions, there will be no errors at startup (unlike with the original version), but git-link-dispatch will give an error unless transient is already installed. (I'm not keen on that, but if people also use magit they won't have an issue).
> 
> `git-link-dispatch` would only be called if they wanted to use the transient menu, right? If so I think this is okay and is sorta typical Emacs behavior. As long as non git-link transient users are not affected.
> 

Correct. Unless people explicitly try to use `git-link-dispatch`, they won't notice any difference. If you're happy with this approach, this PR should be good to merge now.

> > Add transient to Package-Requires. This would require increasing the minimum supported Emacs version to 26.1.
> 
> I try to keep dependencies at a minimum though supporting 26 in 2024/25 does not seem too crazy. But I think point 1 is better approach. Though why 26 if transient was ~adding~ added in 28?

Transient only supports 26+, so on 26 and 27 it can be installed from ELPA. I haven't tested what happens if you try to use it on older versions though.

PR #134, opened 2025-01-17
Optionally translate Git hosts to web hosts

sebastianpoeplau:
Add an option to translate Git hosts to web hosts in order to support setups where the two are different. For example, Git access could be via "ssh.gitlab.company.com" and the web interface at "gitlab.company.com".


sshaw:
Hi, thanks so much for the PR. I'm surprised this issue has not come up befor. https://github.com/sshaw/git-link/issues/32 is the closest I could find.

git-link.el
```diff
@@ -845,22 +860,25 @@ With a double prefix argument invert the value of
...
         (message "Remote `%s' not found" remote)
 
       (setq remote-info (git-link--parse-remote remote-url)
+            git-host    (car remote-info)
             filename    (git-link--relative-filename)
             branch      (git-link--branch)
             commit      (git-link--commit)
-            handler     (git-link--handler git-link-remote-alist (car remote-info)))
+            handler     (git-link--handler git-link-remote-alist git-host)
+            web-host    (or (assoc-default git-host git-link-web-host-alist) git-host))
```
sshaw, line 874: We can use `assoc` since we're using `or` to provide a default; _or_, provide the default to `assoc-default`. I think the rest of code uses the `or` pattern so maybe that is a better choice?

sebastianpoeplau: The problem with `assoc` is that it returns the pair `(key . value)`, so that we'd have `(cdr (assoc ...))`. No problem if that's what you prefer, but the extra `cdr` is why I didn't use it. As for `assoc-default`, unfortunately it respects the default value only for keys that are in the list but not associated with any value. For example:

```elisp
(assoc-default 'apple '((apple . red) pear) nil 'unknown) ;; -> red
(assoc-default 'pear '((apple . red) pear) nil 'unknown) ;; -> unknown
(assoc-default 'orange '((apple . red) pear) nil 'unknown) ;; -> nil, not unknown as we'd expect...
```

sshaw: Ah yes I see. Thanks for clarifying.

Looks good then, just one question: the other host-related alists do a regex match on the host, should we do that here for consistency?

sshaw:
Great, thanks so much!

PR #136, opened 2025-03-07
Use git-link-web-host-alist in git-link-commit and git-link-homepage

sebastianpoeplau:
Links to commits or project homepages now also use the optional translation mechanism from Git hosts to web hosts.

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -588,6 +588,14 @@ return (FILENAME . REVISION) otherwise nil."
   link
   )
 
+(defun git-link--web-host (git-host)
+  "Determine the web host to use for GIT-HOST.
+
+The translation is based on `git-link-web-host-alist'. If there
+is no entry for GIT-HOST in the list, it is returned unmodified."
+  (or (assoc-default git-host git-link-web-host-alist #'string-match-p)
+      git-host))
+
 (defun git-link-codeberg (hostname dirname filename branch commit start end)
     (format "https://%s/%s/src/%s/%s"
 	    hostname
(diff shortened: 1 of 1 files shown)


sshaw:
Thanks for the PR. Totally missed that #134 only applied to `git-link`!

Is it possible t combine this web-host mapping into a function that can be shared across the 3 uses? High-level it looks like it but, I did not dig in.

sebastianpoeplau:
> Is it possible t combine this web-host mapping into a function that can be shared across the 3 uses? High-level it looks like it but, I did not dig in.

Absolutely. I actually thought about it but then forgot :see_no_evil:

sebastianpoeplau:
I extracted the code into a new function `git-link--web-host` but never mentioned here that I did...

sshaw:
Great thanks. Will give it a whirl later today

git-link.el
```diff
@@ -917,24 +924,27 @@ With a prefix argument prompt for the remote's name.
 Defaults to \"origin\"."
 
   (interactive (list (git-link--select-remote)))
-  (let* (commit handler remote-info (remote-url (git-link--remote-url remote)))
+  (let* ((remote-url (git-link--remote-url remote))
+         commit handler remote-info git-host web-host)
```
sshaw, line 928: Any reason for reversing these?

sebastianpoeplau: If I remember correctly, when I added `git-host` and `web-host` the line became very long and looked cleaner when split. Then it seemed more idiomatic to place the variables without an initial value at the end, but I don't think it was necessary in any way :thinking:

PR #142, opened 2025-06-23
Allow + in URL schemas for git links

sid-code:
If you clone a repository using `git+ssh://` the parser won't match it and you can't get a link.


sshaw:
Thanks for the fix! Surprised this did not come up before. 

Can you [add a test for this](https://github.com/sshaw/git-link/blob/8d0f98cf36f6b9c31285329b054ae77f9a3d9b33/git-link-test.el)?

PR #147, opened 2025-08-11
Codeberg handler. format permalink correctly.

mooseyboots:
Codeberg permalinks add `/commit/` after `/src` and before the commit hash. This PR makes `git-link` create correct permalinks to code references for Codeberg.

Currently, if you visit a `git-link` created commit URL in the browser, Forgejo/Codeberg will re-write it to add the `/commit/` part. I'm working on an emacs client for Forgejo (https://codeberg.org/martianh/fj.el), so I can't benefit from such re-writing and need `git-link` to create the right kind of permalink.


sshaw:
Great thanks! Left a note for reference but this is good to go!

PR #151, opened 2026-01-18
If in magit-revision buffer, git-link can open the commit

FrostyX:
I think this better than printing an error that

> Can't figure out what to link to

When in a `magit-revision` buffer, people most likely want to link the displayed commit. But this also allows to link a commit that the cursor is pointing to.

### DIFF
--- a/git-link.el
+++ b/git-link.el
@@ -951,15 +951,20 @@ Defaults to \"origin\"."
     (if (null remote-url)
         (message "Remote `%s' not found" remote)
 
+      (setq commit (word-at-point))
+      (when (and (not (string-match-p "[a-fA-F0-9]\\{7,40\\}" (or commit "")))
+                 (derived-mode-p 'magit-revision-mode)
+                 (functionp 'magit-buffer-revision))
+        (setq commit (magit-buffer-revision)))
+
       (setq remote-info (git-link--parse-remote remote-url)
             git-host (car remote-info)
-            commit (word-at-point)
             handler (git-link--handler git-link-commit-remote-alist git-host)
             web-host (git-link--web-host git-host))
 
       (cond ((null git-host)
              (message "Remote `%s' contains an unsupported URL" remote))
-            ((not (string-match-p "[a-fA-F0-9]\\{7,40\\}" (or commit "")))
+            ((not commit)
              (message "Point is not on a commit hash"))
             ((not (functionp handler))
              (message "No handler for %s" git-host))
--- a/README.md
+++ b/README.md
@@ -163,8 +163,9 @@ For example:
 <img src="./magit-revision-buffer.png" width="400" alt="magit-revision buffer demo">
 
 Here one can get a link to:
-- The commit by calling `git-link-commit` on a commit hash
 - A file by calling `git-link` on a filename or within its diff
+- The commit by calling `git-link-commit` on a commit hash or anywhere in the
+  buffer (besides filename or its diff)
 
 ### Building Links and Adding Services
 


FrostyX:
Considering your screenshot ... If I install a `git-link` from `master` and move the cursor to the exact place you have, then:

- `M-x git-link` will result in `Can’t figure out what to link to`
- `M-x git-link-commit` will result in `Point is not on a commit hash`

Which on one hand makes sense and basically tells me that I need to move my cursor on some commit hash and run `M-x git-link-commit` again. On the other hand, it was quite obvious what I wanted to do. I am in a `magit-revision`, I ran `git-link-commit`, so I clearly wanted a link for that very commit.

In the PR I am improving `git-link-commit` to support this. At the same time, it is only a fallback. If the cursor points to some hash, it will open that hash. The new behavior only applies if the cursor doesn't point to a hash and `git-link-commit` would previously result in an error.

> and links to the files (file in this case) by calling git-link on their name or diff.

I did not know that, thank you. That's a nice feature and my PR accidentally breaks that. What I was trying to achieve was ... if in a `magit-revision` buffer with a cursor pointing either to a commit hash or being anywhere in the commit header section, I wanted `git-link` to behave like `git-link-commit` and open the commit, instead of erroring out.

sshaw:
Added a Magit Revision Buffer section to the docs d9b375f79e6071a9926bf73bba64111adfc93bf5

FrostyX:
> Added a Magit Revision Buffer section to the docs https://github.com/sshaw/git-link/commit/d9b375f79e6071a9926bf73bba64111adfc93bf5

Nice, I think it will be helpful :-)

> Great, thanks for clarifying. I wanted to understand what is going on high-level before lookin' at the code (too much! :neckbeard:).

My apologies, when writing the comment I realized I didn't explain the change in the PR description clearly enough.

> If so then that could fix the current breakage that you refer to.

I think the change was almost correct, I just improved the condition. Please take a look, I think it works exactly how we want now.

sshaw:
Thanks for the update. 

Is there a need [to modify `git-link`](https://github.com/sshaw/git-link/pull/151/changes#diff-a86ea92b0e180064b421a637b78c7d5ae152de93c60a82136b77a6a72c90b35cR907)? From your description it sounds like we only need to modify `git-link-commit`

FrostyX:
It wasn't necessary. I thought of it as a nice helper, allowing to just use `git-link` instead of having to remember `git-link-commit` but it is not what this PR was really about, so I dropped it :-)

git-link.el
```diff
@@ -951,15 +951,19 @@ Defaults to \"origin\"."
     (if (null remote-url)
         (message "Remote `%s' not found" remote)
 
+      (setq commit (word-at-point))
+      (when (and (not (string-match-p "[a-fA-F0-9]\\{7,40\\}" (or commit "")))
+                 (derived-mode-p 'magit-revision-mode))
+        (setq commit (magit-buffer-revision)))
```
sshaw, line 958: This function does not exist for me: 

```
setq: Symbol’s function definition is void: magit-buffer-revision
```

I'm on Emacs 29.4 with Magit 20241001.2052. We should make sure it works on old versions (assuming this is the problem)

sshaw: > I'm on Emacs 29.4 with Magit 20241001.2052. We should make sure it works on old versions (assuming this is the problem)

If this function was introduced in later versions and there's no alternative we can call `(functionp)` for the time being to guard against cases where it does not exist.

FrostyX: Good catch @sshaw, sorry I missed that.
I didn't want to complicate the code too much because of this niche feature, so I just added the guard clause as you suggested.

sshaw: Does using the `magit-buffer-revision` **variable** work for you locally? I see that this is what was used before the function version was introduced in 8-2025. If that works for then what do you think about checking for it first? 

Maybe there's a case to use the variable always? The function in newer so I assume it's better long-term but that may not be true.

git-link.el
```diff
@@ -951,15 +951,20 @@ Defaults to \"origin\"."
...
+
       (setq remote-info (git-link--parse-remote remote-url)
             git-host (car remote-info)
-            commit (word-at-point)
             handler (git-link--handler git-link-commit-remote-alist git-host)
             web-host (git-link--web-host git-host))
 
       (cond ((null git-host)
              (message "Remote `%s' contains an unsupported URL" remote))
-            ((not (string-match-p "[a-fA-F0-9]\\{7,40\\}" (or commit "")))
```
sshaw, line 962: When in non-Magit revision buffers, removing this results in links being generated to non commits (or as best we can tell given the regex).

sshaw:
Fixed in b651de4

FrostyX:
Thank you very much for finishing the PR @sshaw, I think it is a really nice feature, but I couldn't find the time to take another look.

Thank you again <3

PR #152, opened 2026-04-28
fix: git-link-homepage-remote-alist missing codeberg.org

Kyrremann:
I noticed that the homepage-alist is missing `codeberg.org`. Not sure of the order, but did the same as the other two alists.

I use git-link togther with link-hint to quickly open my repo page in Magit status buffer.
``` elisp
'(link-hint-action-fallback-commands
	 '(:open
		 (lambda nil
			 (condition-case _
					 (progn (browse-url (git-link-homepage "origin")) t)
				 (error nil)))))```


sshaw:
Thanks! Have not heard of link-hint, will check it out.

PR #153, opened 2026-06-11
Add `--no-show-signature' to git log args.

sadboy:
When the `git log` output contains GPG signature info (which can happen e.g. when the user's `.gitconfig` has `showSignature = true`) it confuses the revision parser. Here we add an arg to `git log` to explicitly disable any GPG signature info.


sshaw:
Thanks for the PR!

PR #155, opened 2026-07-23
Correct custom variable type attributes

syohex:
`repeat` type should be used for them because those values are variable length string list. While `list` type is used for fixed length list.

This also fixes byte-compile warnings

```
git-link.el:320:10: Warning: in defcustom for
    ‘git-link-extensions-rendered-plain’: ‘list’ without arguments
git-link.el:329:10: Warning: in defcustom for
    ‘git-link-extensions-rendered-via-bitbucket-annotate’: ‘list’ without arguments
```

#### Reference
- https://www.gnu.org/software/emacs/manual/html_node/elisp/Composite-Types.html


sshaw:
Thanks for the PR. 

What Emacs version do you see these warnings on?

syohex:
> What Emacs version do you see these warnings on?

Development version. 

```
$ emacs --version
GNU Emacs 32.0.50
Development version b31954283556 on master branch; build date 2026-07-06.
```

sshaw:
Great, thanks!

